PR #2, opened 2014-12-24
Travis fixes

swails:
This seems to have the desired effect for me on my machine.

Note that 2 of the test comparisons are currently failing, one of them because it seems like a test file was mistakenly omitted from the commit.  One of the lines from the `Test_Error.dat` output is:

```
  nofit.dat.save not found.
```

Also: `displ.dat.save displ.dat are different.`, and the differences look pretty serious...

This will fail the Travis build (or should fail it, anyway), but the problems should be fixed in a different PR.  These are only github-specific changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,4 +10,4 @@ install:
     - make install
 script:
     - cd test && make test
-    - cat test/Test_Error.dat
+    - cat Test_Error.dat
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,4 +10,3 @@ install:
     - make install
 script:
     - cd test && make test
-    - cat Test_Error.dat
--- a/configure
+++ b/configure
@@ -8,7 +8,7 @@
 
 # Print help message
 Usage() {
-  echo "Usage: ./configure [gnu | intel | pgi] OPTIONS"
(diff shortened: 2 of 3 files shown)


mojyt:
I'll add the missing test file. The modes displacement fail is actually interesting - the sign of certain (not all) eigenvectors are flipped which is why the displacement comes out different but the fluctuations do not. Since the sign of an eigenvector is arbitrary this doesn't necessarily seem bad, but it also doesn't seem great that you can get different results depending on what BLAS/LAPACK version you have. I'll have to look into this a bit more but I may disable this test for now.

swails:
If the sign doesn't matter, why not always use the absolute value?  Or am I missing something obvious?

mojyt:
You can't just use the absolute value of vector components because then all vectors become restricted to a certain subspace. For example, if one eigenvector is {1, 1, 0} and my second eigenvector is {-1, 1, 0}, if I take the absolute value of the components they become the same (which is incorrect). I'm actually thinking of just removing this functionality completely. I don't see how it is useful or why it was put there in the first place by Holger and Alrun. I also don't remember ever seeing this kind of analysis in the literature.

PR #11, opened 2015-06-04
CPPTRAJ HN: add PyTrajInfo for cpptraj to return a vector of string

hainm:
to get info about how cpptraj was compiled.


hainm:
@mojyt 
Dan,

what do you think about this? `pytraj` might need to have the info here to have 'good' installation. For example, if `cpptraj` is compiled with `openmp` and `pytraj` is not, there is segmentation fault for some Actions. So my idea is: before compile whole pytraj files, I just compile one to get general info and change the compiler flag according to.

Hai

src/Cpptraj.h
```diff
@@ -1,15 +1,21 @@
...
  *  overall flow of the program. It exists in main.cpp.
  */
+using std::vector;
+using std::string;
+
 class Cpptraj {
   public:
     Cpptraj() {}
     int RunCpptraj(int, char**);
+    static vector<string> ForPytrajInfo();
```
hainm: need to change to better name.

mojyt:
Hi,

Do you need this to be a vector of strings (i.e. is that easiest for you to
process?). Or could you parse a single string? Either way, I would like to
just have this function called "Defines()" and then have the '--defines'
command line flag use that as well. This way if new defines are added only
one function needs to be updated.

-Dan

On Wed, Jun 3, 2015 at 11:46 PM, Hai Nguyen <notifications@github.com>
wrote:

> @mojyt <https://github.com/mojyt>
> Dan,
>
> what do you think about this? pytraj might need to have the info here to
> have 'good' installation. For example, if cpptraj is compiled with openmp
> and pytraj is not, there is segmentation fault for some Actions. So my
> idea is: before compile whole pytraj files, I just compile one to get
> general info and change the compiler flag according to.
>
> Hai
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mojyt/cpptraj/pull/11#issuecomment-108735914>.
>



-- 
-------------------------
Daniel R. Roe, PhD
Department of Medicinal Chemistry
University of Utah
30 South 2000 East, Room 307
Salt Lake City, UT 84112-5820
http://home.chpc.utah.edu/~cheatham/
(801) 587-9652
(801) 585-6208 (Fax)

hainm:
@mojyt a single string is fine with me, as long as `pytraj` can take the output from cpptraj, not the stdout.

I agree about Defines() one. I will update the code and let you review again. thx.

Hai

hainm:
@mojyt I just pushed new change.

PR #13, opened 2015-06-09
CPPTRAJ HN: add Box& in Topology to inplace update Topology.box in pytraj

hainm:
Motivation: with current implementation of `cpptraj`, `top.box` in `pytraj` will return a copy of Box.
If user accidentally uses `top.box[0] = 100.`, `top.box` is not updated.

`pytraj` code
```python
    property box:
        def __get__(self):
            cdef Box box = Box()
            box.thisptr[0] = self.thisptr.ParmBox()
            return box
```

Proposed new code in `pytraj` (with PR for `cpptraj` too)
```python
    property box:
        def __get__(self):
            cdef Box box = Box()
            box.thisptr = &self.thisptr.ParmBox()
            box.py_free_mem = False # create a `view`
            return box
```

**Note**: `thisptr` is pointer to cpptraj's Box object

### DIFF
--- a/src/CoordinateInfo.h
+++ b/src/CoordinateInfo.h
@@ -18,6 +18,7 @@ class CoordinateInfo {
       remdDim_(r), box_(b), ensembleSize_(e), hasVel_(v), hasTemp_(t), hasTime_(m), hasFrc_(f) {}
     bool HasBox()              const { return box_.HasBox();            }
     const Box& TrajBox()       const { return box_;                     }
+    Box& TrajBox()                   { return box_;                     }
     int EnsembleSize()         const { return ensembleSize_;            }
     bool HasVel()              const { return hasVel_;                  }
     bool HasTemp()             const { return hasTemp_;                 }
--- a/src/Frame.h
+++ b/src/Frame.h
@@ -85,6 +85,7 @@ class Frame {
     double Mass(int atnum)            const { return Mass_[atnum];   }
     /// \return Box information
     const Box& BoxCrd()               const { return box_;           }
+    Box&  BoxCrd()                          { return box_;           }
     RemdIdxType const& RemdIndices()  const { return remd_indices_;  }
     // Routines for accessing internal data pointers
     inline double* xAddress() { return X_;                }
--- a/src/Topology.h
+++ b/src/Topology.h
@@ -123,6 +123,7 @@ class Topology {
     int PrintChargeMassInfo(std::string const&, int) const;
     // ----- Routines to Access/Modify Box info --
     inline Box const& ParmBox()   const { return coordInfo_.TrajBox();        }
+    inline Box& ParmBox()               { return coordInfo_.TrajBox();        }
     inline Box::BoxType BoxType() const { return coordInfo_.TrajBox().Type(); }
     void SetParmBox( Box const& bIn )   { coordInfo_.SetBox( bIn );           }
     // ----- Setup routines ----------------------


hainm:
pytraj' test case

https://github.com/hainm/pytraj/blob/887f387d76afd0536d2198cc29bf1ab7cf74132d/tests/test_Topology_box_modify.py#L10

hainm:
pytraj's test case for Frame

https://github.com/hainm/pytraj/blob/top_box/tests/test_Frame_box_modify.py#L10

mojyt:
Is this the same issue as the other PRs? The TrajBox() routine returns const reference, not a copy. If I'm not understanding the issue correctly though let me know.

hainm:
@mojyt No, this is different issue (box is small, no it's fine to return a copy).

Since TrajBox() return `const&`, `pytraj` can not change its data. (for example. `traj.top.box[0] = 100`.

mojyt:
The way I originally envisioned it working is that the TrajBox() routine returns a const reference (no copying) for cases where you want to get box info, and the SetParmBox(Box const&) routine is used when you want to change Topology box information. See e.g. Action_Box. Does that not work for you? This in principle makes it harder to accidentally invalidate box info (like setting one of the angles negative or something weird like this) although it is of course still possible to do.

hainm:
Hi,
If this make harder to debug the code, I am good with current implementation.

mojyt:
Doesn't necessarily make it harder, but it does mean you can't always be sure of what version of the operator you are getting.

hainm:
yeah, agree.

swails:
The current implementation is more consistent with the overall cpptraj design, and should probably stay.  @hainm -- if you want to make this "pythonic", you should wrap it in your own Python class using the `property` decorator to make it behave the way you want.

hainm:
@swails I already wrap it here

https://github.com/pytraj/pytraj/blob/master/pytraj/core/Box.pyx#L153

I just try to look at `pytraj` as user's perspective.

Currently, when calling `traj.top.box`, a Box copy must be returned (due to `const&` in cpptraj). So assigning `traj.top.box.x = 100` does not update `traj.top.box`.

```python
In [2]: traj.top.box
Out[2]: <Box: ortho, (x, y, z, alpha, beta, gamma) = 
(39.455974, 46.821517, 40.469541, 90.0, 90.0, 90.0)>

In [3]: traj.top.box.x = 100.

In [4]: traj.top.box.x
Out[4]: 39.455974
```

In order to change box info with current implementation, `pytraj` need to do.

```python
In [7]: box = traj.top.box

In [8]: box.x = 100.

In [9]: top = traj.top

In [10]: top.box = box

In [11]: traj.top = top

In [12]: traj.top.box
Out[12]: <Box: ortho, (x, y, z, alpha, beta, gamma) = (100.0, 46.821517, 40.469541, 90.0, 90.0, 90.0)>
```

This is incredibly long since `pytraj` can not inplace update Topology and Box (due to const&).

hainm:
This code does not update `box.y` for `traj.top` with current implementation.

```python
In [17]: box.y = 200.

In [18]: traj.top.box = box

In [19]: traj.top.box.y
Out[19]: 46.821517
```

mojyt:
I guess I just don't understand the way the python interface does things. In C++, the const& is not a copy but a reference - does python not have the equivalent?
Anyway, the point is that if you need to modify the box directly you can create a routine, just avoid overloading the const routine, e.g.
traj.ModifyParmBox().X = 100
or something. Make sense? I know it seems a little pedantic, but I find tracking down issues is much harder when you have subtle overloads like const vs non-const. (Actually tracking down issues is harder with any overloading, but at least if arguments differ that's a more obvious change).

hainm:
>  In C++, the const& is not a copy but a reference - does python not have the equivalent?

if using `const&`, I can not change the content through a pointer, right? `pytraj` use pointer to cpptraj's class, so whenever having `const&`, `pytraj` can not do anything.

>  I know it seems a little pedantic 
:D yeah, too much work for simple task.

What about creating separate methods, like "_get_box()" to return "Box&", "_get_parm()" to return &Topology for only `pytraj`.

PR #19, opened 2015-07-27
hn: update travis link

hainm:


drroe:
Unless I'm mistaken this appears to be the same thing I did with
#17 
How was it possible to have two separate PR do the same thing? And how do we avoid this in the future?

hainm:
Look carefully, bro :p

Hai

drroe:
Mea culpa - I missed a mojyt (as you pointed out). Maybe I need glasses...

PR #22, opened 2015-07-28
add more constructor for Frame

hainm:
@drroe 

So I "officially" create this PR to add one more constructor for Frame. (as we discussed in g-doc a while ago). 

**Motivation**: current implementation of cpptraj require coordinates copied twice when transferring from disk to in-memory numpy array. `cpptraj` read coords from disk to Frame class, then pytraj copy frame's coordinates to numpy (two copies need to be done to convert `discontiguous memory` (vector of frames) to contiguous memory block in numpy).

* I think it's not ok to do this for very large system (may be > 100K atoms).  So I add this constructor to use Frame just as a pointer to a **pre-allocated** numpy array and instruct cpptraj to read coords directly to numpy memory block. Memory is not own by cpptraj but numpy (controlled through `pytraj`).

So this constructor is just designed for cpptraj/pytraj interface.
This implementation 2 times speeds up + memory saving.

* Secondly, this is also speed up the frame slicing of numpy array. Current implementation needs to slice numpy array, convert to Frame object so cpptraj can understand. If we use constructor above, no data is copied. Of course, this only good for cpptraj's actions that does not change mem location. 

let me know if it's ok to merge.

Hai

### DIFF
--- a/src/Frame.cpp
+++ b/src/Frame.cpp
@@ -56,6 +56,20 @@ Frame::Frame(int natomIn) :
     X_ = new double[ ncoord_ ];
 }
 
+Frame::Frame(int natomIn, double* ptr) :
+  natom_(natomIn),
+  maxnatom_(natomIn),
+  ncoord_(natomIn*3), 
+  T_(0.0),
+  time_(0.0),
+  X_(0),
+  V_(0),
+  Mass_(natomIn, 1.0)
+{
+  if (ncoord_ > 0)
+    X_ = ptr;
+}
+
 // CONSTRUCTOR
 Frame::Frame(std::vector<Atom> const& atoms) :
   natom_(atoms.size()),
--- a/src/Frame.cpp
+++ b/src/Frame.cpp
@@ -25,6 +25,7 @@ const size_t Frame::COORDSIZE_ = 3 * sizeof(double);
 // ---------- CONSTRUCTION/DESTRUCTION/ASSIGNMENT ------------------------------
 /// CONSTRUCTOR
 Frame::Frame( ) : 
+  only_free_V(false),
   natom_(0),
   maxnatom_(0),
   ncoord_(0),
@@ -37,12 +38,18 @@ Frame::Frame( ) :
 /// DESTRUCTOR
 // Defined since this class can be inherited.
 Frame::~Frame( ) { 
-  if (X_ != 0) delete[] X_;
-  if (V_ != 0) delete[] V_;
+  if (only_free_V) {
+        if (V_ != 0) delete[] V_;
+   }
+  else {
+    if (X_ != 0) delete[] X_;
+    if (V_ != 0) delete[] V_;
+  }
 }
 
 // CONSTRUCTOR
 Frame::Frame(int natomIn) :
+  only_free_V(false),
   natom_(natomIn),
   maxnatom_(natomIn),
   ncoord_(natomIn*3), 
@@ -57,6 +64,7 @@ Frame::Frame(int natomIn) :
 }
 
 Frame::Frame(int natomIn, double* ptr) :
+  only_free_V(true),
   natom_(natomIn),
   maxnatom_(natomIn),
   ncoord_(natomIn*3), 
@@ -72,6 +80,7 @@ Frame::Frame(int natomIn, double* ptr) :
 
 // CONSTRUCTOR
 Frame::Frame(std::vector<Atom> const& atoms) :
+  only_free_V(false),
   natom_(atoms.size()),
   maxnatom_(natom_),
   ncoord_(natom_*3),
@@ -90,6 +99,7 @@ Frame::Frame(std::vector<Atom> const& atoms) :
 
 // CONSTRUCTOR
 Frame::Frame(Frame const& frameIn, AtomMask const& maskIn) : 
+  only_free_V(false),
   natom_( maskIn.Nselected() ),
   maxnatom_(natom_),
   ncoord_(natom_*3),
@@ -131,6 +141,7 @@ Frame::Frame(Frame const& frameIn, AtomMask const& maskIn) :
 
 // COPY CONSTRUCTOR
 Frame::Frame(const Frame& rhs) :
+  only_free_V(false),
   natom_(rhs.natom_),
   maxnatom_(rhs.maxnatom_),
   ncoord_(rhs.ncoord_),
--- a/src/Frame.cpp
+++ b/src/Frame.cpp
@@ -25,7 +25,7 @@ const size_t Frame::COORDSIZE_ = 3 * sizeof(double);
 // ---------- CONSTRUCTION/DESTRUCTION/ASSIGNMENT ------------------------------
 /// CONSTRUCTOR
 Frame::Frame( ) : 
-  only_free_V(false),
+  OnlyFreeV_(false),
   natom_(0),
   maxnatom_(0),
   ncoord_(0),
@@ -38,7 +38,7 @@ Frame::Frame( ) :
 /// DESTRUCTOR
 // Defined since this class can be inherited.
 Frame::~Frame( ) { 
-  if (only_free_V) {
+  if (OnlyFreeV_) {
         if (V_ != 0) delete[] V_;
    }
   else {
@@ -49,7 +49,7 @@ Frame::~Frame( ) {
 
 // CONSTRUCTOR
 Frame::Frame(int natomIn) :
-  only_free_V(false),
+  OnlyFreeV_(false),
   natom_(natomIn),
   maxnatom_(natomIn),
   ncoord_(natomIn*3), 
@@ -64,7 +64,7 @@ Frame::Frame(int natomIn) :
 }
 
 Frame::Frame(int natomIn, double* ptr) :
-  only_free_V(true),
+  OnlyFreeV_(true),
   natom_(natomIn),
   maxnatom_(natomIn),
   ncoord_(natomIn*3), 
@@ -80,7 +80,7 @@ Frame::Frame(int natomIn, double* ptr) :
 
 // CONSTRUCTOR
 Frame::Frame(std::vector<Atom> const& atoms) :
-  only_free_V(false),
+  OnlyFreeV_(false),
   natom_(atoms.size()),
   maxnatom_(natom_),
   ncoord_(natom_*3),
@@ -99,7 +99,7 @@ Frame::Frame(std::vector<Atom> const& atoms) :
 
 // CONSTRUCTOR
 Frame::Frame(Frame const& frameIn, AtomMask const& maskIn) : 
-  only_free_V(false),
+  OnlyFreeV_(false),
   natom_( maskIn.Nselected() ),
   maxnatom_(natom_),
   ncoord_(natom_*3),
@@ -141,7 +141,7 @@ Frame::Frame(Frame const& frameIn, AtomMask const& maskIn) :
 
 // COPY CONSTRUCTOR
 Frame::Frame(const Frame& rhs) :
-  only_free_V(false),
+  OnlyFreeV_(false),
   natom_(rhs.natom_),
   maxnatom_(rhs.maxnatom_),
   ncoord_(rhs.ncoord_),
--- a/src/Frame.h
+++ b/src/Frame.h
@@ -196,7 +196,7 @@ class Frame {
     int RecvFrame(int);
 #   endif
   private:
-    bool only_free_V;
+    bool OnlyFreeV_;
     typedef std::vector<double> Darray;
     static const size_t COORDSIZE_;
     static const size_t BOXSIZE_;
--- a/src/Frame.h
+++ b/src/Frame.h
@@ -196,6 +196,7 @@ class Frame {
     int RecvFrame(int);
 #   endif
   private:
+    bool only_free_V;
     typedef std::vector<double> Darray;
     static const size_t COORDSIZE_;
     static const size_t BOXSIZE_;
--- a/src/Frame.h
+++ b/src/Frame.h
@@ -36,6 +36,8 @@ class Frame {
     ~Frame();
     /// Set up empty frame for given # of atoms.
     Frame(int);
+    /// Set up Frame as a memoryview. mostly for pytraj.
+    Frame(int, double*);
     /// Set up to be the size of given atom array (including masses).
     Frame(std::vector<Atom> const&);
     /// Copy input frame according to input mask.


hainm:
just ignore the 1st "push" since I accidentally pust my commit directly to Amber-MD repo. The 2nd is PR from hainm/cpptraj (pass travis).

swails:
I think we should let Dan control merges for anything of substance to cpptraj.  Likewise, you should do all pytraj merges and I should do all ParmEd merges.

drroe:
@hainm it seems like code using this constructor could run into problems since the Frame destructor is still always called. I'm not sure how this is being used in pytraj but something like:
```C++
    double* MyCoords = new double[30];
    Frame MyFrame(10, MyCoords);
    delete[] MyCoords;
```
will double free. Also, currently if the Frame set up in this manner is assigned (`=` op) you get a full copy - is that desired behavior (it might be)?
If you're not having any memory issues than maybe this is OK, but if not then we can just add a bool that tracks whether this constructor was called and memory does not need to be freed. Thoughts?

swails:
This is a good point.  If the `Frame` is not supposed to "own" the data, then it shouldn't be freed during the `Frame` destructor.

Since I think pytraj is going to pass in a `numpy.ndarray` (or something similar), two destructors will be trying to free that memory and you'll get a double-free (is it true that on some compilers the second delete is ignored? so maybe this will only cause problems on certain compilers?).

> If you're not having any memory issues than maybe this is OK, but if not then we can just add a bool that tracks whether this constructor was called and memory does not need to be freed. Thoughts?

If the intent is a memory view (as the comments suggest), then `Frame` should track whether or not it "owns" certain data and only free the arrays it is told it owns.  So +1 to Dan's idea from me.

hainm:
> I think we should let Dan control merges for anything of substance to cpptraj. Likewise, you should do all pytraj merges and I should do all ParmEd merges.

agree.

@drroe 
I've created new pytraj branch and new cpptraj just to test new constructor and so far I don't see the double-free memory. 

This is how Cython work: To wrap cpptraj's Frame (I named as _Frame), I create a Cython's Frame class. In Frame constructor, pytraj use `new` to create a _Frame object. To avoid memory leak, a destructor need to be call to explicitly deallocate _Frame's memory.

**initialize **
```python
    def __cinit__(self):
        self.thisptr = new _Frame()
```

**cython's destructor**
```python
    def __dealloc__(self):
        if self.py_free_mem and self.thisptr:
            del self.thisptr
```

To use cpptraj's _Frame as a memoryview, I just set the flag `py_free_mem = False` and give the mem owning to numpy.

hainm:
> Also, currently if the Frame set up in this manner is assigned (= op) you get a full copy - is that desired behavior (it might be)?

Current Cython version I used does not allow '=' op, cython does its own '=' so I did not know '=' return a full copy for Frame or not. But I rarely use '=', just explicitly make `copy` method or use `frame = Frame(frame0)` to have a copy`.

swails:
Either you're leaking memory or you're double-freeing still.  You can't avoid both of these strictly from Python.  From the look of it, the way you're addressing this is by disabling the C++ `Frame` destructor from ever running, just leaving a dangling pointer that goes out of scope but is never deallocated.

This is a leak.  If you deallocate the bulk of the used memory elsewhere, then the leak is small(er), but it's still there.

If you *are* calling the `Frame` destructor, then you're clearly double-freeing.

hainm:
@swails you're absolutely right. I've checked memory and there is memory leak.


Original implemtation of Frame constructor (my 1st PR)
```python
Filename: traj_iter_new_frame_init.py

Line #    Mem usage    Increment   Line Contents
================================================
    15     56.7 MiB      0.0 MiB   @profile
    16                             def new_iter():
    17    468.4 MiB    411.7 MiB       traj = pt.io._load_netcdf("remd.x.000", "myparm.top")
    18    600.2 MiB    131.8 MiB       for f in traj._iter_new_constructor():
    19    600.2 MiB      0.0 MiB           pass
    20    733.5 MiB    133.2 MiB       for f in traj._iter_new_constructor():
    21    733.5 MiB      0.0 MiB           pass
    22    866.7 MiB    133.2 MiB       for f in traj._iter_new_constructor():
    23    866.7 MiB      0.0 MiB           pass
```

I've look at cpptraj's code [here](https://github.com/Amber-MD/cpptraj/blob/master/src/Frame.cpp#L37-L42) and it turn out that the allocation for velocity  has never been freed.

So now I add `only_free_V` [flag] (https://github.com/hainm/cpptraj/commit/a4a3ad40eb64667a7a2745ddc9687d6372994440)  in cpptraj code and let Cython always deallocate Frame object.
Here is the new memory profile.

```python
Filename: traj_iter_new_frame_init.py

Line #    Mem usage    Increment   Line Contents
================================================
    15     56.7 MiB      0.0 MiB   @profile
    16                             def new_iter():
    17    468.4 MiB    411.7 MiB       traj = pt.io._load_netcdf("remd.x.000", "myparm.top")
    18    468.6 MiB      0.2 MiB       for f in traj._iter_new_constructor():
    19    468.6 MiB      0.0 MiB           pass
    20    468.6 MiB      0.0 MiB       for f in traj._iter_new_constructor():
    21    468.6 MiB      0.0 MiB           pass
    22    468.6 MiB      0.0 MiB       for f in traj._iter_new_constructor():
    23    468.6 MiB      0.0 MiB           pass
```

You can see that there is no new memory appear

hainm:
add some cpptraj's actions

```python
Filename: traj_iter_new_frame_init.py

Line #    Mem usage    Increment   Line Contents
================================================
    15     56.7 MiB      0.0 MiB   @profile
    16                             def new_iter():
    17    468.4 MiB    411.7 MiB       traj = pt.io._load_netcdf("remd.x.000", "myparm.top")
    18    468.6 MiB      0.2 MiB       for f in traj._iter_new_constructor():
    19    468.6 MiB      0.0 MiB           pass
    20    468.6 MiB      0.0 MiB       for f in traj._iter_new_constructor():
    21    468.6 MiB      0.0 MiB           pass
    22    468.6 MiB      0.0 MiB       for f in traj._iter_new_constructor():
    23    468.6 MiB      0.0 MiB           pass
    24
    25    472.2 MiB      3.6 MiB       pt.molsurf(traj, '@CA')
    26    472.9 MiB      0.7 MiB       pt.radgyr(traj)
    27    473.1 MiB      0.1 MiB       pt.multidihedral(traj, '!:WAT')
```

swails:
> You can see that there is no new memory appear

Note that the memory resolution you're looking at here is not necessarily going to suffice to find small leaks.  For instance, consider the following class:

```C++
class SomeClass {

public:
    SomeClass::SomeClass(int something, double *external_array) : var1_(something) {
        array_ = external_array;
    }

private:
    size_t var1_;
    double *array_;
}
```

Now suppose you do something like this:

```C++
{
    double stuff[100];
    SomeClass *foo = new SomeClass(0);
}
// stuff goes out of scope, so foo->array_ has been freed from the stack.
// But you have still leaked sizeof(SomeClass)
```

In this case, you've leaked only a couple bytes of memory, but it's a leak nonetheless.  If this happens enough times, that tiny leak can become arbitrarily large.

Although I *don't* think that's what's happening here.  In pure C++, valgrind will be able to report leaks of any size.  You should be able to do that with Python, too, although you may have to suppress false errors from valgrind for that.

hainm:
yeah, you're right about `sizeof(SomeClass)`, so with my new commit, I always let cython free memomry for Frame class itself.

https://github.com/hainm/pytraj/blob/Frame_pt_constructor/pytraj/Frame.pyx#L126-L132
(by setting `py_free_mem` = True)

hainm:
and I just googled `valgrind python`, seems to be complicated to dig.

drroe:
@hainm This still isn't really addressing the heart of the matter: if Frame doesn't own the data it should *not* attempt to free it, i.e. if memory allocation occurs outside of Frame, destruction should as well. I think it's better to do something like this:
```C++
Frame::Frame(int natomIn, const double* xIn) : 
  natom_(natomIn),
  maxnatom_(natomIn),
  ncoord_(natomIn*3), 
  T_(0.0),
  time_(0.0),
  X_(xIn),
  V_(0),
  Mass_(natomIn, 1.0),
  freeMemory_(false)
{ ... }

Frame::~Frame() {
  if (freeMemory_) {
    if (X_ != 0) delete[] X_;
    if (V_ != 0) delete[] V_;
  }
}
```
Then the other constructors get freeMemory_ set to true. Make sure you update the copy constructor, swap function, and assignment operator as well. All the SetupFrame... routines should probably be protected as well. This is a semi-large change to the intent of Frame so we need to be careful.

hainm:
@drroe: yeah, I already added the flag for memory free (similiar to your code). Note that only Frame `xyz` coordinates is own by others (`numpy`), so I still let `cpptraj` free `V_`.

https://github.com/Amber-MD/cpptraj/pull/22/files

but I have not done for swap and copy. 

I agree that we should be careful about this if you think it will break cpptraj. I will close this PR and play with my own cpptraj branch.

swails:
> @drroe: yeah, I already added the flag for memory free (similiar to your code). Note that only Frame xyz coordinates is own by others (numpy), so I still let cpptraj free V_.

I think it would be cleaner if you always grouped `X_` and `V_` together.  If you allocate `X_` outside of `Frame`, then allocate `V_` outside of `Frame`, too (or just pass in `NULL` if you don't need it).

Then you change your `OnlyFreeV_` to `freeMemory_` and you arrive at Dan's solution.  It's weird for the `Frame` destructor to free either all of its arrays or only some of them.  Better to do all or none IMO.

hainm:
you are right, @swails

hainm:
@drroe 

is there any test case I can run to make sure that my PR (constructor) does not break cpptraj?

(currently this PR passed all provided cpptraj's tests)

swails:
It won't. If it's a new overloaded operator, then by definition no place in cpptraj is using it. 

Since you're playing with how memory is allocated and deallocated, I would suggest running a valgrind test to make sure there are no leaks.

hainm:
thanks. I will try to see if it's easy to use valgrind with python.

swails:
No i meant use valgrind for the built-in cpptraj tests. If you want to make sure your commit didn't break anything in cpptraj, all you have to do is make sure you didn't introduce any leaks.

hainm:
I see.

drroe:
@hainm I see this is marked closed right now. Do you want me to take a crack at implementing this?

As Jason mentioned, current tests (either the ones here or my extended tests) wouldn't show any problems with this because no part of the code would be using this constructor. This is an API-design issue - we want to make sure that future programmers (us and maybe others) get "expected behavior" when they use this code. The convention is that if a class creates memory, that class should be responsible for freeing it (and vice versa). If you still want to work on this that's fine, but if not let me know and I'll be happy to implement this (it's been on my ToDo list since we first talked about it anyway).

hainm:
@drroe 

Yes, if it's not a big problem with you, you're perfect candidate to implement it (so I can focus more on pytraj).  My PR is a demonstration for my idea.

I think this is a good feature since both cpptraj and numpy can operate with the same data without copy back and forth (fast numpy vectorization + fast frame iterating (since everything is just moving pointer)).

swails:
This PR is already close IMO.  The only thing that's missing is for the allocation of `V_` to be done in concert with `X_` and to rename `OnlyFreeV_` to `ownsMemory_`.  Right?

src/Frame.cpp
```diff
@@ -25,6 +25,7 @@ const size_t Frame::COORDSIZE_ = 3 * sizeof(double);
 // ---------- CONSTRUCTION/DESTRUCTION/ASSIGNMENT ------------------------------
 /// CONSTRUCTOR
 Frame::Frame( ) : 
+  OnlyFreeV_(false),
```
swails:
```c++
ownsMemory_(false),
```

hainm:
@swails: no, I have not introduced flags to `swap` and `copy` methods yet.

src/Frame.cpp
```diff
@@ -37,12 +38,18 @@ Frame::Frame( ) :
...
-  if (X_ != 0) delete[] X_;
-  if (V_ != 0) delete[] V_;
+  if (OnlyFreeV_) {
+        if (V_ != 0) delete[] V_;
+   }
+  else {
+    if (X_ != 0) delete[] X_;
+    if (V_ != 0) delete[] V_;
+  }
 }
```
swails:
```c++
  if (ownsMemory_) {
    if (X_ != 0) delete[] X_;
    if (V_ != 0) delete[] V_;
  }
```

src/Frame.cpp
hainm: good name.

I know from now, I should always come to your desk and ask a name. lol.

src/Frame.cpp
```diff
@@ -37,12 +38,18 @@ Frame::Frame( ) :
...
 
 // CONSTRUCTOR
 Frame::Frame(int natomIn) :
+  OnlyFreeV_(false),
```
swails:
```c++
ownsMemory_
```

src/Frame.cpp
```diff
@@ -56,8 +63,24 @@ Frame::Frame(int natomIn) :
...
+  time_(0.0),
+  X_(0),
+  V_(0),
+  Mass_(natomIn, 1.0)
+{
+  if (ncoord_ > 0)
+    X_ = ptr;
+}
+
 // CONSTRUCTOR
```
swails: Something like:
```c++
Frame::Frame(int natomIn, double* crd, double *vel) :
  ownsMemory_(true),
  natom_(natomIn),
  maxnatom_(natomIn),
  ncoord_(natomIn*3), 
  T_(0.0),
  time_(0.0),
  X_(0),
  V_(0),
  Mass_(natomIn, 1.0)
{
  X_ = crd;
  V_ = vel;
}
```

src/Frame.h
```diff
@@ -36,6 +36,8 @@ class Frame {
     ~Frame();
     /// Set up empty frame for given # of atoms.
     Frame(int);
+    /// Set up Frame as a memoryview. mostly for pytraj.
+    Frame(int, double*);
```
swails:
```c++
Frame(int, double*, double*);
```

src/Frame.h
```diff
@@ -194,6 +196,7 @@ class Frame {
     int RecvFrame(int);
 #   endif
   private:
+    bool OnlyFreeV_;
```
swails:
```C++
bool ownsMemory_;
```

swails:
You get the idea.  @hainm -- you could get this PR fixed up well pretty quickly IMO -- most of the work is already done. Just some variable renames and minor changes.

hainm:
I just want to "spam" a bit to add one more reason why I need this feature.

since `parmed` require `numpy`, it makes sense to have pytraj to use numpy too. Currently, `pytraj` has 3 trajectory classes: TrajectoryIterator (our-of-core traj, just like cpptraj's idea), Trajectory (vector of Frame: was designed to avoid extra copy to Frame when using cpptraj's action. The drawback of this class is that users need to copy to numpy array), and api.Trajectory (which is similar to `mdtraj`, using numpy as backend; the drawback of this is that pytraj need to make data copy to Frame object so cpptraj can understand).

If the feature is implemented, I just need to combine `Trajectory` and `api.Trajectory` and use numpy as backend + cpptraj's many actions.

hainm:
> You get the idea. @hainm -- you could get this PR fixed up well pretty quickly IMO -- most of the work is already done. Just some variable renames and minor changes.

OK  Jason. @drroe so I guess just let me fix and PR again.

src/Frame.h
hainm: @swails 

what's usage of `..double*, double*)`? for V_ too?

If I don't want to allocate V_, just use NULL, right?

drroe:
@swails @hainm For this solution to really be complete all of the routines that allocate memory probably should be looked at as well, since now e.g. if you call SetupFrameV() after constructing with external memory you will get a double free.
The way I have the Frame class manging memory is not necessarily optimal - maybe I should revisit it...

hainm:
ah, yes. You should do it. ^^

src/Frame.h
swails: Yea, that's the idea.

swails:
Ah, I didn't realize there were other places of allocating.

PR #29, opened 2015-07-31
DRR - Cpptraj: Add ability to average data sets together.

drroe:
This PR adds functionality to the 'avg' analysis that lets one calculate the average (and SD) over sets as opposed to the average of each individual set.


hainm:
@drroe does this new feature have a test yet?

drroe:
It does in my private repo. It will take a bit before I get those tests over to the GitHub repo (there are 109 tests comprising 404 comparisons total). I need to go through and consolidate where I can to make it more compact.

hainm:
Cool

Hai

PR #47, opened 2015-08-18
Add support for reduce-generated PDB files.

swails:
@hainm -- see if this does the trick...


hainm:
ok. I will test it.

drroe:
Thanks Jason.
Ugh, they couldn't just use one of the existing like 50 or so tags? Had to come up with a non-standard one?

hainm:
I just tried both cpptraj and parmed. They works with old reduce now. thanks.

swails:
> Ugh, they couldn't just use one of the existing like 50 or so tags? Had to come up with a non-standard one?

I know, right?  It wasn't actually the parsing that didn't work, it was the format auto-detection...

PR #56, opened 2015-09-01
Frame external data

drroe:
This creates a constructor and associated functions that allows external coordinate data to be assigned to Frame; current primary use is in pytraj.


hainm:
cool. I've been waiting for this.

PR #57, opened 2015-09-02
Reorganize trajectory classes.

drroe:
This is a significant reorganization of the classes that handle reading and writing trajectories at the highest level: Trajin, Trajin_Single, Trajin_Multi, etc. Previously, there was not a consistent interface for the trajectory classes. For example, ensemble functionality was merged along with functionality to read in single frames in the Trajin_Multi class.

Now there are two types of input trajectories with no mixing of functionality: the Trajin base class (Trajin_Single, Trajin_Multi) for reading 1 frame in at a time, and the EnsembleIn base class (EnsembleIn_Single, EnsembleIn_Multi) for reading in an array of frames at a time. Similarly there are two types of output trajectories: the EnsembleOut (EnsembleOut_Single, EnsembleOut_Multi) base class and the Trajout_Single class (not a base class since currently there is no need to write a single frame out to multiple files as this is already handled by TrajoutList). Some helper classes have also been created with common functionality: InputTrajCommon, TrajFrameCounter, and TrajIOarray for input; OutputTrajCommon for output. Another change is that the input and output classes no longer share TrajectoryFile as a common base class; instead this has become a static class which is used under the hood for TrajectoryIO class allocation.


hainm:
@drroe does this PR only change internally? for example, does this add more syntax to traj.in? 

If yes, will you add examples?

hainm:
:D just want to confirm that this PR breaks pytraj. have work to do now.

```
pytraj/trajs/Trajin.cpp:11925:28: error: ‘class Trajin’ has no member named ‘SetTrajParm’
pytraj/trajs/Trajin.cpp:11932:58: error: ‘class Trajin’ has no member named ‘TrajParm’
error: command 'gcc' failed with exit status 1
```

hainm:
I just updated pytraj. Things are good now.

PR #63, opened 2015-09-09
Convert to using Container-based Travis-CI infrastructure

swails:
Faster builds (~1/2 the time), more resources (2 dedicated CPUs instead of 1.5
burst; 4 GB RAM instead of 3.5 GB). Builds also start almost instantly, rather
than sometimes needing to wait.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,5 +20,7 @@ env:
 install:
     - ./configure $OPENMP $COMPILER
     - make -j2 install
+      # Hack to get ambpdb installed with OpenMP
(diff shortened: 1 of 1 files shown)


swails:
This is ready to merge.  Building with clang emits *so* many warnings, so I disabled that (since it made the log file too large so Travis killed it).  At some point it might be nice to get that working, but for now let's skip it.

This also makes sure to test both the OpenMP version (on 2 cores) and the serial version.

drroe:
Looks outstanding - thanks Jason

PR #65, opened 2015-09-11
Major overhaul of DataSet/DataFile framework, file name handling

drroe:
In this PR the DataSet framework is given a major overhaul. There are far too many changes to completely list but here are some of the highlights.

The biggest change is that the variables used to categorize and describe DataSets (e.g. name, aspect, index, etc) have been consolidated into a single class called MetaData. This not only makes sense conceptually, but will avoid major API changes if/when new meta data is added down the road (which was the case when ensembleNum was introduced). One result is that the numerous versions of AddSet in DataSetList (AddSetIdx, AddSetAspect, etc) have been reduced to only 3.

DataSets can now have a file name associated with them; this is searchable meta data. This has allowed simplification of the reference frame framework in DataSetList and has allowed the reference frame functionality to be better integrated. This necessitated fixing up the FileName class; all functionality related to file names is now in FileName instead of spread out between that and StringRoutines. All functions that used a string to convey a file name now uses FileName. This is more sensible and fixes issues like tilde expansion not working for output trajectories etc.

Similar to the MetaData change, all data in DataSet that was associated with formatting the text output (width, precision etc) has been encapsulated in the TextFormat class.

Reading in data sets from data files has also been improved; the logic of whether to append or not or whether double data sets need to be converted to mesh data sets (for non-monotonic X values) is now completely within DataSetList.


drroe:
Forgot to mention that this PR also introduces the concept of AssociatedData to DataSets. This is data that is related to the data set but not necessarily relevant for all sets of that type. So for example instead of making NOE-related variables a part of DataSet_double, it becomes an AssociatedData class which can be added to whatever DataSet.

PR #66, opened 2015-09-15
add Datasets to Action_PairDist

hainm:
Currently Hannes only dumps data to file, so I added 3 Datasets to this action.

Let me know what you think. I can update other Actions if needed.

### DIFF
--- a/src/Action_PairDist.cpp
+++ b/src/Action_PairDist.cpp
@@ -4,7 +4,7 @@
 #include "CpptrajStdio.h"
 #include "StringRoutines.h"
 #include "DistRoutines.h"
-
+#include "DataSet_double.h"
 
 
 /** Calculate pair distribution function P(r) between two masks.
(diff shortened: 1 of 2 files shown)


drroe:
Looks fine - however, since this is really X-Y data (distance, P(r)) it might be better to have two XYMESH data sets instead of three DOUBLE and use DataSet_Mesh::AddXY(dist, Pr) etc. This way if you want to e.g. integrate the data internally it will use the correct X values. If you would like to change it let me know. Otherwise I can just merge this PR now and implement the DataSet_Mesh in a separate PR in the future.

hainm:
@drroe: what's about 'std_' dataset? keep it in 'DataSet_double'? I have no problem with updating the code.

hainm:
oh, you meant two Mesh

`DataSet_Mesh::AddXY(dist, Pr)`
`DataSet_Mesh::AddXY(dist, std)`

right?

drroe:
> oh, you meant two Mesh
> DataSet_Mesh::AddXY(dist, Pr)
> DataSet_Mesh::AddXY(dist, std)
> right?

Yes exactly. This way the X values for the data sets will be correct internally. You could still use DataSet_double if the X values are guaranteed to be monotonic, but the `if (Pr > 0.0)` statement means that some X values may not have defined Pr values, so I think DataSet_Mesh is a better choice in this case. Make sense?

hainm:
oop, wrong branch.

hainm:
@drroe the new PR is here: https://github.com/Amber-MD/cpptraj/pull/67

PR #67, opened 2015-09-15
Pairdist dataset

hainm:


hainm:
@drroe 
sorry to ask dump question: any idea how to produce this data from my PR? :D  
 
https://github.com/Amber-MD/cpptraj/blob/master/test/Test_PairDist/Pr.dat.save

drroe:
You can't exactly because of the `# pair-distance distribution P(r)` title, which is something that the  standard data file format doesn't support at the moment. Otherwise you should be able to get this using the standard data output format, setting the legends to match the columns (`P(r)` and `stddev`), then changing the X dimension label to `#distance`. You will probably also need to modify the default precision. However, I wouldn't worry about matching that output format exactly - what matters is that the numbers are the same. If we decide to go entirely DataSet/DataFile with this action we can update the test output.

drroe:
@hainm also, since a header file was added can you update the dependencies as well:
```
cd $CPPTRAJHOME/src
make depend
```
Then add the updated cpptrajdepend. Thanks!

hainm:
> Then add the updated cpptrajdepend. Thanks!

can you update cpptraj guide doc for this? thanks. (I just updated my PR).

hainm:
I was not clear about reproducing data. The number from DataSetList in CpptrajState does not match the save output number.

It's clear that I did not understand Mesh well.

a part of data from CpptrajState for distance vs Pr.

```
[[0, 30.0],
 [0, 1030.0],
 [0, 150.0],
 [0, 327.62376237623766],
 [0, 423.16831683168306],
 [0, 363.3663366336633],
 [0, 95.0495049504951],
 [0, 173.26732673267324],
 [0, 83.46534653465348],
 [0, 137.02970297029705],
 [1, 676.9306930693068],
 [1, 1224.9504950495048],
 [1, 544.4554455445544],
 [1, 558.6138613861385],
 [1, 1196.6336633663366],
 [1, 934.1584158415841],
 [1, 705.2475247524749],
```

this is from saved file
```
#distance P(r) stddev
    0.9500            30.00       0.00
    1.0500          1030.00       0.00
    1.2500           150.00       0.00
    1.3500           327.62       1.58
    1.4500           423.17       1.65
    1.5500           363.37       1.31
    1.6500            95.05       2.24
    1.7500           173.27       2.43
    1.8500            83.47       2.39
    1.9500           137.03       3.26
    2.0500           676.93       6.39
    2.1500          1224.95       7.74
    2.2500           544.46       5.33
    2.3500           558.61       5.98
    2.4500          1196.63       7.55
```

hainm:
oh boy, the distance info is not kept in MESH data? I need to check ``pytraj`` code to make sure it takes the data from cpptraj correctly.

drroe:
> oh boy, the distance info is not kept in MESH data? 

Based on your code here:
```
+      ((DataSet_Mesh*) Pr_)->AddXY(dist, Pr);
+      ((DataSet_Mesh*) std_)->AddXY(dist, sd);
```
The `dist` variable is being stored correctly (in DataSet_Mesh::mesh_x_). When Cpptraj writes out or needs to use a coordinate variable from a DataSet (i.e. X for a 1-dimensional DataSet, X or Y for a 2-dimensional DataSet etc) it uses the DataSet::Coord() function. This ensures that the correct value for the coordinate in question is used (it takes into account X dimension min and step values for DataSet_double for example). DataSet_1D has a specific function for the X coordinate called Xcrd(). I suspect that you are just printing the index for X in the above output. Does this make sense?

hainm:
:d my fault in ``pytraj``. I convert to ``int`` for X value in MESH to be used for residue number in bfactor calculation. Stupid me. :((

hainm:
figure_1

hainm:
@drroe can you merge now? thanks.

PR #79, opened 2015-09-23
Diffusion code change

hainm:
This is not complete update code for Action_STFC_Diffusion.cpp.

Just want to make this PR for discussion. Currently I used DataSet_double for storing `avgx, avgy, avgz and avg` values for each frames.

When ``pytraj`` takes the data, it needs to reshape the data set to (n_frames-1, 4)

What's your suggestion?


hainm:
happy merging. we have openmp now.

PR #99, opened 2015-10-02
make code consistent: use DSL and DFL

hainm:
'datasetlist' --> 'DSL' (to be consistent with all Action class)
'DFLin' --> "DFL"

### DIFF
--- a/src/AnalysisList.cpp
+++ b/src/AnalysisList.cpp
@@ -31,11 +31,11 @@ void AnalysisList::SetDebug(int debugIn) {
   * DataSetList.
   */
 int AnalysisList::AddAnalysis(DispatchObject::DispatchAllocatorType Alloc, ArgList& argIn,
-                              DataSetList* DSLin, DataFileList* DFLin)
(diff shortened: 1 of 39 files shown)


hainm:
@drroe let me know what do you think.

swails:
`DSLin` -> `DSL` and `DFLin` -> `DFL` are probably bad changes.  The former names are examples of Hungarian notation done correctly.  `datasetlist` -> `DSL` probably improves consistency and are good changes.

@drroe will have stronger opinions than I (and stronger justification as well), but that's my "more-outsider's" opinion.

hainm:
@swails 

after playing with cpptraj code for a while, it seems to me that cpptraj uses only a single DataSetList to store everything (input (add topology, reference, dataset, ...) and output).

But I agree that @drroe will have stronger opinion. This PR update is done by auto-replacing without carefully reading the code. :D

hainm:
uhm, no longer need this PR (check https://github.com/Amber-MD/cpptraj/pull/111)

PR #106, opened 2015-10-09
No -lgfortran for clang

rmcgibbo:
Sorry, my editor cleans up whitespace automatically (cleaner version of the [diff](https://github.com/Amber-MD/cpptraj/pull/106/files?w=true)). The one change is to remove `-lgfortran` from `FLIBS` when using `clang`, since OS X comes with Accelerate framework for blas/lapack that doesn't require linking against libgfortran.

### DIFF
--- a/configure
+++ b/configure
@@ -310,7 +310,7 @@ while [[ ! -z $1 ]] ; do
       OMPFLAGS=""
       FFLAGS="-ffree-form"
       FOPTFLAGS="-O3"
-      FLIBS="$FLIBS -w"
+      FLIBS="$FLIBS -lgfortran -w"
       PICFLAG="-fPIC"
       ;;
     "intel" )
@@ -660,6 +660,14 @@ if [[ $USEMKL -eq 1 ]] ; then
   fi
   BLAS=""
   LAPACK=""
+fi
+
+if [[ "$(uname)" == "Darwin" && "$FFT_LIB" == "-lfftw3" ]] ; then
+    # Linking against fortran libraries (e.g. -lgfortran) is not required on
+    # Mac OS X using FFTW3, since the the Mac Accelerate blas/lapack/arpack
+    # doesn't require any extra fortran libs, and fftw3 does not require
+    # fortran.
+    FLIBS=""
 fi
 
 # Set up linking flags
--- a/configure
+++ b/configure
@@ -131,16 +131,16 @@ TestMathlib() {
   if [[ ! -z $ARPACK ]] ; then # Assume BLAS and LAPACK also defined
     cat > testp.cpp <<EOF
 #include <cstdio>
-extern "C" { 
+extern "C" {
   void dsaupd_(int&, char&, int&, char*, int&, double&, double*,
                int&, double*, int&, int*, int*, double*, double*,
                int&, int&);
 }
 int main() {
-  int ival = 0; 
+  int ival = 0;
   double dval = 0.0;
-  char cval = 'I'; 
-  dsaupd_(ival, cval, ival, &cval, ival, dval, &dval, 
+  char cval = 'I';
+  dsaupd_(ival, cval, ival, &cval, ival, dval, &dval,
           ival, &dval, ival, &ival, &ival, &dval, &dval,
           ival, ival);
   printf("Testing\n"); return 0;
@@ -166,7 +166,7 @@ EOF
   fi
 }
 
-# Test C/C++/Fortran compilers 
+# Test C/C++/Fortran compilers
 TestCompile() {
   # C
   echo "Testing C compiler:"
@@ -174,7 +174,7 @@ TestCompile() {
 #include <stdio.h>
 int main() { printf("Testing\n"); return 0; }
 EOF
-  $CC $CFLAGS -o testp testp.c > /dev/null 2> compile.err 
+  $CC $CFLAGS -o testp testp.c > /dev/null 2> compile.err
   ./testp | grep "Testing" > /dev/null
   status=$?
   if [[ $status -gt 0 ]] ; then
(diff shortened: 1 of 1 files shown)


swails:
What if I want to use clang on my Linux box?  Ideally we should leave `-lgfortran` in-place unless the kernel is Darwin, or is that already done here?

drroe:
I guess the configure script should try and determine whether the build environment is OS X or not and then remove the `-lgfortran` flag when the compiler is clang - does that sound reasonable?

swails:
Yes.

drroe:
Seems like we can borrow this code from configure2 for OSX detection:
```sh
if [ `uname -s|awk '{print $1}'` = "Darwin" ]; then
```

swails:
Yep.

rmcgibbo:
I saw that there was a flag for MKL'S BLAS. Perhaps the better way is to
add a flag for Accelerate, and have the link line here set based on the
BLAS, not the compiler per se.

swails:
And fft implementation I think. @drroe-- isn't pubfft the default fft implementation used? (Which is written in Fortran). 

--
Jason M. Swails
BioMaPS,
Rutgers University
Postdoctoral Researcher

rmcgibbo:
Oh, is that it? I've been using `-fftw`, so if it's only the fft lib that requines linking against libgfortran, then it may simply be keyed to the compiler option when it should be keyed to the fft option.

swails:
It's both the math libraries and the fft library that might depend on Fortran cross-compiling. So it needs to be keyed to both. Macs without fftw still need -lgfortran

drroe:
Currently `-lgfortran` is required when linking with the built-in pub_fft or with fortran versions of BLAS/LAPACK/ARPACK (which is what comes bundled with Amber). Since CPPTRAJ originally used math libraries and pub_fft from Amber (all fortran), the `-lgfortran` flag was always included. Perhaps now a new variable in the configure script should be created (something like FORTRAN_LIB) which can be set to the appropriate flag when linking to fortran code is required.

rmcgibbo:
Okay, I updated this to only remove `-lgfortran` for OS X when using `-fftw3`.

swails:
+1 for this

swails:
@rmcgibbo -- how did you get this configured to use -framework Accelerate instead of the "standard" math libraries?

I don't see an option for it, and trying to configure with clang on my Mac gives me:

```
[swails@Robin ~/src/cpptraj (clang) ]$ ./configure -fftw3 --with-netcdf=/opt/local clang

Using FFTW for FFT.
Using NETCDF in /opt/local
Using clang/gfortran compilers
Installing cpptraj binary to /Users/swails/src/cpptraj/bin
Testing C compiler:
  OK
Testing C++ compiler:
  OK
Checking BZLIB
  OK
Checking ZLIB
  OK
Checking NetCDF
  OK
Checking ARPACK/LAPACK/BLAS
./configure: line 88: ./testp: No such file or directory
  Error: Unable to compile using:
         clang++  -O3 -Weverything  -DFFTW_FFT -DHASBZ2 -DHASGZ -DBINTRAJ -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64  -I/opt/local/include -o testp testp.cpp -larpack -llapack -lblas  
  Please check your compiler settings or configure flags. Error follows:
  --------------------
ld: library not found for -larpack
clang: error: linker command failed with exit code 1 (use -v to see invocation)
  --------------------
```

rmcgibbo:
I have a libarpack.dylib in /usr/local/lib. Maybe from brew or something.

swails:
OK, I have some fixes that get this to work without a libarpack.dylib sitting around in /usr/local.  I'll submit a PR to your branch @rmcgibbo

swails:
Actually, @rmcgibbo -- this doesn't quite work on my machine.  Here are the linking errors I get on my Mac:

```
Undefined symbols for architecture x86_64:
  "_dsaupd_", referenced from:
      DataSet_Modes::CalcEigen(DataSet_2D const&, int) in DataSet_Modes.o
  "_dseupd_", referenced from:
      DataSet_Modes::CalcEigen(DataSet_2D const&, int) in DataSet_Modes.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[1]: *** [cpptraj] Error 1
make: *** [install] Error 2
```

"-framework Accelerate" isn't sufficient, because I think they only provide lapack and blas routines, not arpack...  So unless I'm mistaken, you'll still need `-lgfortran` unless you turn math libraries off completely (i.e., Accelerate is not a drop-in replacement for everything cpptraj needs).

rmcgibbo:
It's not quite that you "need -lgfortran", is that you need arpack, and depending on what arpack you use and how it was built, it may or may not require other libraries like libgfortan.

swails:
How can we determine that programmatically?  Because on my Mac, I need -lgfortran...

rmcgibbo:
This is really what pkgconfig is for.

drroe:
Where are we on this? Should this wait until ARPACK dependencies are split from BLAS/LAPACK?

swails:
No idea :)

rmcgibbo:
I think the patch is still worth merging. It's relatively non dangerous -- just resetting FLIBS on OS X and linking against fftw for fft.

swails:
Do we not need the `-framework Accelerate` CL flags?

rmcgibbo:
No, `-lblas` and `-llapack` are sufficient.
```
$ ll /usr/lib/libblas.dylib
lrwxr-xr-x  1 root  wheel   116B Dec  7  2014 /usr/lib/libblas.dylib -> ../../System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
```

swails:
Ah, got it.

drroe:
> I think the patch is still worth merging. It's relatively non dangerous -- just resetting FLIBS on OS X and linking against fftw for fft.

Sounds good - thanks for the work!

Issue #105, opened 2015-10-09
hbond with image: got 0 


hainm:
I was trying to use image on/off but got 0 hbond with image on

**case 1**
```bash
parm tz2.ortho.parm7
trajin tz2.ortho.nc
hbond out hb.dat
```

output
```bash

#Frame   HB_00001[UU]
       1            5
       2            5
       3            5
       4            4
       5            4
       6            5
       7            6
       8            5
       9            5
      10            4
```

**case 2**
```bash
parm tz2.ortho.parm7
trajin tz2.ortho.nc
hbond image out hb.dat
```

output
```bash
#Frame   HB_00001[UU]
       1            0
       2            0
       3            0
       4            0
       5            0
       6            0
       7            0
       8            0
       9            0
      10            0
```

according to this topic http://archive.ambermd.org/201506/0455.html, is 'image' in hbond similiar to `autoimage`? if yes, why the output in two cases are so different?

hainm:
full log for case 2

```

CPPTRAJ: Trajectory Analysis. V16.00b
    ___  ___  ___  ___
     | \/ | \/ | \/ |
    _|_/\_|_/\_|_/\_|_

| Date/time: 10/09/15 00:20:15
| Available memory: 508.012 MB

INPUT: Reading Input from file tz2.hbond.in
  [parm tz2.ortho.parm7]
        Reading 'tz2.ortho.parm7' as Amber Topology
  [trajin tz2.ortho.nc]
        Reading 'tz2.ortho.nc' as Amber NetCDF
  [hbond image out hb.dat]
  HBOND: Searching for Hbond donors/acceptors in region specified by *
        Distance cutoff = 3.000, Angle Cutoff = 135.000
        Writing # Hbond v time results to hb.dat
        Imaging enabled.
---------- RUN BEGIN -------------------------------------------------

PARAMETER FILES:
 0: 'tz2.ortho.parm7', 5293 atoms, 1704 res, box: Orthogonal, 1692 mol, 1691 solvent, 10 frames

INPUT TRAJECTORIES:
 0: 'tz2.ortho.nc' is a NetCDF AMBER trajectory, Parm tz2.ortho.parm7 (Orthogonal box) (reading 10 of 10)
  Coordinate processing will occur on 10 frames.
TIME: Run Initialization took 0.0000 seconds.

BEGIN TRAJECTORY PROCESSING:
.....................................................
[...]

PR #107, opened 2015-10-09
Fix Hbond imaging in orthorhombic systems

drroe:
Imaging in hbond always requires calculation of Cart.->Frac./Frac.->Cart. matrices, so ensure they are calculated when imaging is on.


drroe:
Addresses #105

Issue #93, opened 2015-09-30
Error: Output file is required [out <filename>]


hainm:
I am not if we chat about this.

What's you opinion about strictly requiring output filename?
If the Action is used in pytraj, I prefer to have the filename as an option.

drroe:
Much of this is from legacy ptraj code, where if you didn't supply an output file the command didn't make sense because the data was not stored. Make a list of Actions/Analyses you find with this behavior and post them here. I'll get to them in time.

hainm:
Okie

hainm:
@drroe 

this is the list so far

* Action_AtomicCorr.cpp
* Action_Watershell.cpp
* Action_Radial.cpp
* Action_PairDist.cpp
* Action_Density.cpp
* Action_Grid.cpp

hainm:
* Analysis_Rotdif.cpp

hainm:
* Action_Bounds.cpp: if filename is not provided, the output will be STDOUT.

hainm:
* Action_NativeContacts.cpp: STDOUT

hainm:
* Action_MRT.cpp

hainm:
* Action_Average.cpp

`.Error: average: No filename given`

hainm:
@drroe I updated checkbox in my 1st post. If you finish any Action, you can check the box. It will show the progress in the issue. quite cool. https://github.com/Amber-MD/cpptraj/issues

hainm:
* Analysis_Lifetime.cpp

if `runanalysis lifetime HB[solutehb] out test.dat` does not have `out test.dat`, got STDOUT

```bash
#Set       Nlifetimes      MaxLT      AvgLT  TotFrames SetName
         0         19          4     1.4211         27 GLU_13@O-VAL_2@N-H
         1          1          1     1.0000          1 ACE_1@O-PHE_3@N-H
         2          6          2     1.1667          7 GLU_13@O-PHE_3@N-H
         3          2          1     1.0000          2 VAL_2@O-ILE_4@N-H
         4         21         12     2.5238         53 TYR_11@O-ILE_4@N-H
         5         16          4     1.6250         26 LYS_9@O-SER_6@N-H
         6          6          1     1.0000          6 TYR_11@O-SER_6@N-H
         7          2          1     1.0000          2 THR_5@O-GLY_8@N-H
         8         12          4     1.4167         17 SER_6@O-LYS_9@N-H
[...]

Issue #82, opened 2015-09-24
"Active reference" not passed to COORDS data sets


drroe:
When a COORDS data set is created it creates a separate copy of the associated topology. However, this means that any selection by distance that relies on the COORDS internal topology reference coordinates will fail since these are not ever updated. It may be time to rework the "internal reference" coords of topology files used for distance-based mask selection.

drroe:
For my info: this is the relevant part of the bug as reported by a user:
```
parm pro_com.prmtop
reference wat_box.inpcrd
trajin ../traj/md1.nc
rmsd rms_test :250<:10&@CA,N,C reference out rms_test.dat time 0.002
Error: No reference set for [pro_com.prmtop], cannot select by distance
Error: Could not initialize action [rmsd]
```

drroe:
Fixed by #111

PR #111, opened 2015-10-09
Topology dataset

drroe:
This PR is a major change to the Action API that will hopefully make it more stable and easier to use and understand. Most passed in variables are now wrapped into classes (ActionInit, ActionSetup, and ActionFrame for Init(), Setup(), and DoAction() respectively). If new variables ever need to be passed in they can be added to these classes instead of changing the definition and implementation of every Action (which is what was happening previously). An additional bonus to this approach is that intent within Actions is more clear; for example, if the Action wants to manipulate the current Frame it has to call ModifyFrm(), etc.

Associated with this change, some important variables that were passed into Actions indirectly via Topology have now been taken out of Topology and put into ActionSetup: CoordinateInfo (the current trajectory coordinate info) and the number of frames expected to be associated with a given Topology. This, as well as being more sensible (since there is no reason a Topology should be associated with a specific CoordinateInfo or number of frames) now allows Actions to do things like add velocity information to the Frame without downstream memory problems.

Actions now also have more return options; notably SKIP (for non-fatal Setup errors), MODIFY_TOPOLOGY (for when an Action has modified Topology/CoordinateInfo), and MODIFY_COORDS (for when an Action has modified or manipulated the Frame). As well as making the intent of the Action more clear, this also means that during `crdaction` the COORDS DataSet is only modified if it needs to be.

Some added bonuses; the changes in this PR have addressed issue #82 as well as a few of the issues in #93. Also, the `combinecrd` command now handles merging residue info correctly, and the `pairwise` cut Mol2 file is now printed with bond info. 

There may be a few hiccups left that I haven't seen but all tests (including those in the extended test suite) pass, so I think it's good to go.


hainm:
Files changed 218: oh boy.

hainm:
> ModifyFrm

I prefer to use 'ModifyFrame`, use 'Frm' interrupt my though flow.

drroe:
> I prefer to use 'ModifyFrame`, use 'Frm' interrupt my though flow.

I use `ModifyFrm()` because the `const&` version is `Frm()`, and I use `Frm()` because 'Frame' is already a class. So both to avoid name clashes and to keep methods consistent, I use `ModifyFrm()`

Issue #92, opened 2015-09-29
'modes trajout' does not correctly detect output format based on extension


drroe:
Found by J. Robertson.
example:
```
readdata evecs.dat
modes name evecs.dat trajout traj.nc tmode 1
```
writes trajectory in amber trajectory format, while
```
readdata evecs.dat
modes name evecs.dat trajout traj.nc trajoutfmt netcdf tmode 1
```
writes netcdf.

drroe:
Fixed by #114

PR #114, opened 2015-10-10
Fix 'modes trajout' trajectory format detection

drroe:
'modes trajout' will now correctly determine traj output format from filename extension if 'trajoutfmt' not specified.


drroe:
This addresses #92

PR #124, opened 2015-10-14
Tweaks so we can start testing clang on Travis.

swails:


swails:
This should be good to go.  It adds a clang build (which seems to run faster than the GCC builds, which is not particularly surprising).

drroe:
@swails Cool, thanks Jason.

Issue #128, opened 2015-10-14
add 'RMSF' as an alias to atomicfluct 


hainm:
from this closed issue https://github.com/Amber-MD/cpptraj/issues/126#event-435654012

drroe:
Fixed via #132

Issue #108, opened 2015-10-09
seperate "ifdefs " for BLAS, LAPACK and ARPACK


hainm:
just summary what we've been discussing on gchat.

Currently there is no way to turn on LARPACK, BLAS separately. If using --with-blas, --with-lapack, cpptraj still require libarpack. My idea is to link cpptraj to numpy library (does not have libarpck). This will make it easier to ship libcpptraj via conda.

swails:
What is arpack *not* needed for the lapack and blas *are*?  (And what is arpack needed for?  Just matrix routines?)

swails:
The more flexibility we add on stuff like this (right now we either have *all* math libs or *none*), test coverage becomes exponentially harder.

hainm:
this is from Dan: "arpack is required when not calculating all eigenvectors, e.g. when # vecs < # potential vecs"

hainm:
for the 'arpack' problem, I don't have conda build for it and don't want user install arpack by themself. It's not problem with Amber user, but just thinking about non-Amber user, who don't mind installing AmberTools.

hainm:
It's great if we have conda build for `ARPACK`. Already google and found none. Seem nobody's interested in?

rmcgibbo:
I think I can conda-build arpack

hainm:
> I think I can conda-build arpack

:+1:

rmcgibbo:
OTOH, in general, unless you have very sparse or structured matrices and specialized matrix-vector operations for them, or matrices so large that they don't fit in memory, my experience is that Arnoldi is not competitive with direct methods, even if you only want 1 eigenvector.

swails:
> OTOH, in general, unless you have very sparse or structured matrices and specialized matrix-vector operations for them, or matrices so large that they don't fit in memory, my experience is that Arnoldi is not competitive with direct methods, even if you only want 1 eigenvector.

This has also been my experience.  Admittedly it was not an exhaustive test (and it was a long time ago), but diagonalizing an MM Hessian (when we were trying it out for MMPBSA.py) took basically the same time for Arnoldi with 1 vector as it did for the full diagonalization using one of the direct subroutines.  Once we got up to a small handful of vectors (maybe 4 or 5?) it was actually quite a bit slower, so we ditched it completely.

hainm:
@drroe 

so do we really need ARPACK? :D

drroe:
> so do we really need ARPACK? :D

[...]

Issue #102, opened 2015-10-04
molsurf: set radii


hainm:
from the doc

`This routine currently requires radius information to be present in the topology file`

Should we add SetRadii for Topology so we can use pdb file?

drroe:
Fixed via #132

hainm:
@drroe what's about arbitrary radii? For example. somebody come up with new radii set and want to test molsurf. Do they need to use ParmEd to update radii, save to parm7 and let cpptraj read the radii?

hainm:
just for example with trpcage pdb file (1L2Y). with your new update, I got

```bash
Warning: Setting topology box information from trajectory.
  [molsurf out test.dat]
    MOLSURF: '*' Probe Radius= 1.400 Ang.
    Using GB radii.
---------- RUN BEGIN -------------------------------------------------

PARAMETER FILES:
 0: 1L2Y.pdb, 304 atoms, 20 res, box: Orthogonal, 1 mol

INPUT TRAJECTORIES:
 0: '1L2Y.pdb' is a PDB file, Parm 1L2Y.pdb (Orthogonal box) (reading 38 of 38)
  Coordinate processing will occur on 38 frames.
TIME: Run Initialization took 0.0000 seconds.

BEGIN TRAJECTORY PROCESSING:
.....................................................
ACTION SETUP FOR PARM '1L2Y.pdb' (1 actions):
  0: [molsurf out test.dat]
    Calculating surface area for 304 atoms.
Warning: Atom 'ASN_1@N' has 0.0 GB radius.
Warning: Atom 'ASN_1@CA' has 0.0 GB radius.
Warning: Atom 'ASN_1@C' has 0.0 GB radius.
Warning: Atom 'ASN_1@O' has 0.0 GB radius.
Warning: Atom 'ASN_1@CB' has 0.0 GB radius.
Warning: Atom 'ASN_1@CG' has 0.0 GB radius.
Warning: Atom 'ASN_1@OD1' has 0.0 GB radius.
```

hainm:
### Segmentation fault

```bash
parm 1L2Y.pdb
trajin 1L2Y.pdb
molsurf out test.dat radii vdw
```

```bash

CPPTRAJ: Trajectory Analysis. V16.00b OpenMP
    ___  ___  ___  ___
     | \/ | \/ | \/ |
    _|_/\_|_/\_|_/\_|_

| Date/time: 10/15/15 17:30:00
| Available memory: 6163.49 MB

[...]

Issue #88, opened 2015-09-28
Should support "chi" as alias for "chin" for backwards compatibility.


drroe:
Recent changes to dihedral code have made it so that "chin" means nucleic chi and "chip" is protein chi. However, in ptraj "chi" just meant nucleic chi, so this should be supported.

hainm:
>  in ptraj "chi" just meant nucleic chi, so this should be supported.

is it a good idea? ptraj itself is not perfect. 'chi' is meant for nuc is not reasonable for me.

hainm:
anyway, we (you) need to break 'ptraj' to let users move on to cpptraj.

drroe:
> is it a good idea? ptraj itself is not perfect. 'chi' is meant for nuc is not reasonable for me.

I agree it's not ideal, but it's not worth breaking users' existing scripts IMO. Also, until recently cpptraj had also supported 'chi' to mean 'chin', so the die has already been cast so to speak.

hainm:
is there any easy way to summarize the deprecated warning in the end of cpptraj log?

 I saw some warning in the middle of the stdout but I think most of them will be ignored by users if the log the quite long.

drroe:
Fixed via #132

PR #132, opened 2015-10-15
molsurf/rmsf/dihedral chi/ARPACK

drroe:
This PR addresses several open issues:

#88 For the `dihedral` action `chi` is now an alias for `chin` to preserve backwards compatibility with ptraj.
#108 Users can now compile with `-noarpack` if desired; this will use the full eigenvector calculation under the hood but only save the specified number (which can get expensive for large numbers of eigenvectors).
#102 Added keyword `[radii {gb | parse | vdw}]` to `molsurf` action to enable different radii sets. Cleaned up `molsurf`.
#128 `rmsd` is now an alias for `atomicfluct`


hainm:
> For the dihedral action chi is now an alias for chin to preserve backwards compatibility with ptraj.

I personally stiff don't like this. But ... it's for backward.

hainm:
> rmsd is now an alias for atomicfluct

do you mean 'RMSF'?

drroe:
> do you mean 'RMSF'?

Yes, rmsf

swails:
> I personally still don't like this. But ... it's for backward.

What about doing wild-card recognition?  So `chi*` selects both `chin` and `chip`?

hainm:
@swails  

> So chi* selects both chin and chip?

:+1:  :+1:

Issue #131, opened 2015-10-14
test lowestcurve


hainm:
my test https://github.com/Amber-MD/cpptraj/issues/110#issuecomment-148229056

hainm:
@drroe I have a question about the command:

```bash
  mprintf("\tpoints <# lowest> [step <stepsize>] <dset0> [<dset1> ...]\n"
          "\t[out <file>] [name <setname>]\n"
          "  Calculate a curve of the average of the # lowest points in bins of stepsize.\n");
```

saying if I have a file with 2 columns: RMSD-EPOT

How can I use cpptraj to calculate the data? (what I meant is: can you give me detail traj.in)

![nfig016](https://cloud.githubusercontent.com/assets/4451957/10527505/cb59985c-735e-11e5-9ae4-0a2d141e79a9.gif)

drroe:
```
readdata RMSD-EPTOT.dat index 1 name MyData
runanalysis lowestcurve MyData points 10 step 0.2 name Lowest
writedata All.agr MyData Lowest
```
Not tested but something like that should work. This will calculate the Y average of the 10 lowest points in bins sized 0.2 in X.

hainm:
@drroe 
```bash
$ ../../cpptraj/bin/cpptraj.OMP -i t.in

CPPTRAJ: Trajectory Analysis. V16.00b OpenMP
    ___  ___  ___  ___
     | \/ | \/ | \/ |
    _|_/\_|_/\_|_/\_|_

| Date/time: 10/15/15 17:53:27
| Available memory: 6162.92 MB

INPUT: Reading Input from file t.in
  [readdata EPOTvsRMSD.dat index 1 name MyData]
        Reading 'EPOTvsRMSD.dat' as Standard Data File with name 'MyData'
        Using column 1 as index column.
        DataFile EPOTvsRMSD.dat has 2 columns, 35809 lines.
  [runanalysis lowestcurve MyData points 10 step 0.2 name Lowest]
Error: Only 1D data sets allowed.
Error: Could not add data sets.
TIME: Total analysis execution time: 0.0001 seconds.
        1 errors encountered reading input.
TIME: Total execution time: 0.0431 seconds.
```

input files
-------------

```bash
$ cat t.in
readdata EPOTvsRMSD.dat index 1 name MyData
runanalysis lowestcurve MyData points 10 step 0.2 name Lowest
writedata All.agr MyData Lowest
```

```bash
$ head EPOTvsRMSD.dat
[...]

PR #135, opened 2015-10-16
Protect Actions that use nonbond info, fix detection of numbers, fix lowestcurve

drroe:
Add code in actions that use non-bonded parameters (Lennard-Jones or vdW radius) that checks if non-bonded parameters are actually present (prevents segfaults). Improve number detection. Also do not assume that X values are in order in `lowestcurve`, which could result in incorrect X bins for XY Mesh data sets. Addresses part of #131 and fixes segfault from #102.


hainm:
@drroe can you update cpptraj version? you broke ActionList api.

drroe:
>  you broke ActionList api.

Actually, I *improved* it. :-)

Fix coming up in #137

hainm:
> , I improved it. :-)

ha ha, fair enough.

PR #140, opened 2015-10-20
Update diffusion action

drroe:
Significant change under the hood to the diffusion action. New syntax is implemented:
```
[{out <filename> | separateout <suffix>}] [time <time per frame>]
	[<mask>] [<set name>] [individual] [diffout <filename>] [nocalc]
```
The old syntax and output format are still supported for backwards compatibility (slight differences in output format are fewer whitespace characters and precision in time when step is 1.0 is 0). Diffusion now makes full use of the DataSet/DataFile framework (so you can now write actual grace files or gnuplot etc). By default, only the average diffusion is calculated - the individual calculation can be turned on with the `individual` keyword. Average diffusion sets are stored in double precision; individual diffusion sets are stored in single precision to save memory. 

Linear regression of MSD vs time and calculation of diffusion constants from this is performed by default (can be turned off with `nocalc`) - the slope, intercept, and correlation of these lines is also saved. A bug in the internal orthorhombic imaging has been fixed.

The new calculation is slightly faster than previous for average diffusion constants, but about 25% slower for individual; this is because writing the data files ends up being a bit slower. Since I think the average calc is the more common one I'm OK with this for now. 

This also has two related changes in other classes. The first is that LinearRegression is now in DataSet_1D instead of DataSet_Mesh so that DOUBLE and FLOAT sets can directly perform regression. Eventually I would like all of those types of functions in DataSet_1D instead of DataSet_Mesh. The second is that there is now a version of AddSet, AddSet_NoCheck, in DataSetList that does not check for existing sets (since the check is expensive). This version should only be used by Actions that are sure there will be no Name conflicts. Eventually may need to improve speed of DataSetList CheckSet().


drroe:
@hainm heads up - I don't think this should break pytraj but it might... let me know.

hainm:
@drroe okie.

hainm:
congrats. We're free from breaking.

hainm:
@drroe per `time <time per frame>`, why don't you specify `tstep` like in IRED, velocityautocorr

diffusion
`time Time step in the trajectory. (1.0 ps)`

velocityautocorr
[tstep <timestep>] Time between frames in ps (default 1.0).

PR #147, opened 2015-10-29
Adding support for Amber Netcdf with forces

ctlee:
Working correctly this time. Passes the current unit tests. Still need to add a test for the forces support


drroe:
In `Traj_AmberNetcdf::readForce()` should `vAddress()` be `fAddress()`?

drroe:
Also, why create `Frame::SetupFrameVF()` as opposed to just using `Frame::SetupFrameV`? The issue is that currently any Action that uses SetupFrameV (like e.g. `strip`) will implicitly get rid of forces - if you just made `SetupFrameV()` deal with forces all of these Actions will correctly keep forces if they are defined in the input Frame. Does that make sense?

ctlee:
Yes you're correct, good catch!

ctlee:
@drroe re the `Frame::SetupFrameVF()`, was a naive design decision based off there being multiple `SetupFrame` prototypes available. I wasn't sure if we wanted to be conservative on allocated memory, albeit I believe that when no velocities or forces are encountered, the memory is freed. I will refactor `Frame::SetupFrameV()` accordingly.

drroe:
>  re the Frame::SetupFrameVF(), was a naive design decision based off there being multiple SetupFrame prototypes available.

Well, I made the original naive "design decision" to have differently named SetupXXX routines. In reality the Frame class is one of the oldest in cpptraj, and it has evolved semi-organically as the needs of the code have changed. I still keep the SetupFrame and SetupFrameM routines because they are convenient in some cases (or maybe too inconvenient to change at the moment), but the SetupFrameV routine is the main one, and currently the most flexible since it makes use of CoordinateInfo. So it just makes sense to me to have SetupFrameV allocate space for forces as well if the passed in CoordinateInfo indicates they should be present.

So in summary the issue here was my confusing API design. Everything looks great now, thanks for all the work!

ctlee:
Oh sorry for the misunderstanding. I meant my naive design decision to add
in a new `SetupFrames` function. Didn't spend enough time looking through
the code, and just hacked a solution together! Makes sense to me to have
`SetupFrameV` be as general as possible, and handle memory later.

Ah good look on the test case, forgot that the pytraj issue was for coords
and velocities only. I will update! Thanks for maintaining cpptraj, saves
me a lot of time everyday.

hainm:
@ctlee just curious, have you tried pytraj with your new commits to cpptraj with both f and v?

ctlee:
Not yet. Do I need to compile a new cpptraj.so?

On Thursday, October 29, 2015, Hai Nguyen <notifications@github.com> wrote:

> @ctlee <https://github.com/ctlee> just curious, have you tried pytraj
> with your new commits to cpptraj with both f and v?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Amber-MD/cpptraj/pull/147#issuecomment-152411282>.
>


-- 
_
Christopher T. Lee M.Sc
Ph.D. Student - Amaro & McCammon Group
Department of Chemistry & Biochemistry
University of California, San Diego

Phone: +1 (858) 534-0956
ctlee@ucsd.edu

hainm:
should be 'libcpptraj.so'

`./configure -shared gnu`
`make libcpptraj`

hainm:
oh, you need to `git pull` to update pytraj too.

ctlee:
Got it, I'll take a proper look tomorrow!

ctlee:
This update works with the latest version of Amber-MD/pytraj#972. Passes the simple tests. 

```
bash-4.2$ python runtests.py simple
start testing. Go to ./tests folder
2.7.10 |Anaconda 2.3.0 (64-bit)| (default, Oct 19 2015, 18:04:42) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)]

('pytraj version = ', '0.1.2.dev12')
('cpptraj version = ', u'V16.00b')
('cpptraj internal version = ', u'V4.2.8')
('cpptraj compiled flag = ', u' -DHASBZ2 -DHASGZ -DBINTRAJ')

try to make all dataset stuff
.try to make all action objects
.try to make all analysis objects
.try to make structure-related objects
.try to load sample data
try to make all action objects
try to make all analysis objects
try to make all dataset stuff
try to make structure-related objects
try to create Trajectory-like objects
other stuff. throw all tests don't belong anywhere else here
OK
..other stuff. throw all tests don't belong anywhere else here
.
----------------------------------------------------------------------
Ran 7 tests in 0.255s

OK

HAPPY COMPUTING
```

hainm:
Cool

PR #153, opened 2015-11-02
Fix potential segfault with volmap

swails:
Also adds a default set of radii to use when parameters are not available in the topology.


drroe:
Cool beans

PR #165, opened 2015-11-05
Sequence align

drroe:
This is an experimental command that I used to do some work aligning structures based on BLAST output, but is not ready for release. It is a "hidden" command that will not appear in the online help.


hainm:
care to post an example here?

drroe:
> care to post an example here?

Don't have one on hand - eventually though. This is part of the reason why it is hidden...

PR #174, opened 2015-11-13
Add tests

drroe:
This PR adds several new tests to the GitHub test suite, including a major overhaul of the MasterTest.sh script, which had a bad case of code bloat. Also includes minor fixes to some help text.


hainm:
nice to see some test cases

- multivector
- more rna and dna tests

Issue #172, opened 2015-11-09
atomiccorr: Internal Error: Attempting to assign to Frame with external memory


hainm:
This is not a fatal error, I still got identical output by either out-of-core trajectory or in-memory traj (with fast frame iteration by setting external memory view).

```python
In [4]: traj 
Out[4]: 
pytraj.TrajectoryIterator, 10 frames: 
Size: 0.001183 (GB)
<Topology: 5293 atoms, 1704 residues, 1692 mols, PBC with box type = ortho>
           

In [5]: # load two frames to memory and use fast Frame iter

In [6]: t0 = traj[[0, 3]]

In [7]: pt.atomiccorr(t0, '@CA')
Internal Error: Attempting to assign to Frame with external memory.
Out[7]: 
array([[ 0.        ,  0.66436392,  0.08319134, ...,  0.62662166,
         0.99535167,  0.84034395],
       [ 0.66436392,  0.        ,  0.79999709, ...,  0.89398563,
         0.71788037,  0.16386618],
       [ 0.08319134,  0.79999709,  0.        , ...,  0.69588429,
         0.15919216, -0.45399991],
       ..., 
       [ 0.62662166,  0.89398563,  0.69588429, ...,  0.        ,
         0.69864297,  0.24450834],
       [ 0.99535167,  0.71788037,  0.15919216, ...,  0.69864297,
         0.        ,  0.80383575],
       [ 0.84034395,  0.16386618, -0.45399991, ...,  0.24450834,
         0.80383575,  0.        ]], dtype=float32)

In [8]: # use out-of-core traj, loada single Frame at a time

In [9]: pt.atomiccorr(traj, '@CA', frame_indices=[0, 3])
Out[9]: 
array([[ 0.        ,  0.66436392,  0.08319134, ...,  0.62662166,
         0.99535167,  0.84034395],
       [ 0.66436392,  0.        ,  0.79999709, ...,  0.89398563,
         0.71788037,  0.16386618],
       [ 0.08319134,  0.79999709,  0.        , ...,  0.69588429,
         0.15919216, -0.45399991],
       ..., 
       [ 0.62662166,  0.89398563,  0.69588429, ...,  0.        ,
         0.69864297,  0.24450834],
       [ 0.99535167,  0.71788037,  0.15919216, ...,  0.69864297,
         0.        ,  0.80383575],
       [ 0.84034395,  0.16386618, -0.45399991, ...,  0.24450834,
[...]

PR #176, opened 2015-11-16
Implement assignment of Frame with external memory to frame with internal memory.

drroe:
Should address #172


hainm:
cool. will try.

Issue #178, opened 2015-11-17
clustering: std::bad_alloc. Need nicer message


hainm:
- file: http://www.amber.utah.edu/AMBER-workshop/London-2015/Cluster/ (yours)

```bash
$ cat cluster.in 
parm rGACC.nowat.parm7
trajin rGACC.MREMD1.nowat.nc.40
cluster kmeans clusters 50 @P
```

- kicked
```
ANALYSIS: Performing 1 analyses:
  0: [cluster kmeans clusters 50 @P]
	Starting clustering.
	Mask [@P] corresponds to 3 atoms.
	Calculating pair-wise distances.
	Estimated pair-wise matrix memory usage: > 13407.8309 MB
terminate called after throwing an instance of 'std::bad_alloc'
  what():  St9bad_alloc

Program received signal SIGABRT, Aborted.

```

- trace
```bash
(gdb) bt
#0  0x00000036a1630265 in raise () from /lib64/libc.so.6
#1  0x00000036a1631d10 in abort () from /lib64/libc.so.6
#2  0x00000036aa0bed14 in __gnu_cxx::__verbose_terminate_handler() () from /usr/lib64/libstdc++.so.6
#3  0x00000036aa0bce16 in ?? () from /usr/lib64/libstdc++.so.6
#4  0x00000036aa0bce43 in std::terminate() () from /usr/lib64/libstdc++.so.6
#5  0x00000036aa0bcf2a in __cxa_throw () from /usr/lib64/libstdc++.so.6
#6  0x00000036aa0bd239 in operator new(unsigned long) () from /usr/lib64/libstdc++.so.6
#7  0x00000036aa0bd2f9 in operator new[](unsigned long) () from /usr/lib64/libstdc++.so.6
#8  0x0000000000542046 in Matrix<float>::resize ()
#9  0x00000000005427cb in ClusterMatrix::SetupWithSieve(unsigned long, unsigned long, int) ()
#10 0x000000000053ef6d in ClusterList::CalcFrameDistances(std::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::vector<DataSet*, std::allocator<DataSet*> > const&, ClusterList::DistModeType, ClusterList::DistMetricType, bool, bool, std::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int, int) ()
#11 0x00000000004daf08 in Analysis_Clustering::Analyze() ()
[...]

PR #181, opened 2015-11-17
Improve some output and error messages

drroe:
Some improvements to output messages and error handling. All (I think) messages that wrote out memory sizes have now been standardized by use of the new `ByteString()` function in `StringRoutines.cpp`, which takes a size in bytes and outputs a human-readable size string. When an error occurs during a run, do not attempt to write out data files unless `noexitonerror` has been specified. When exiting cpptraj with an error status, print a message to that effect so it is clear to the user that they need to check for error messages. Also, since it is a likely place for a memory error, catch out-of-memory exceptions when allocating the pair-wise matrix during clustering and print a message that recommends the `sieve` option (addresses #178).


src/StringRoutines.cpp
```diff
@@ -215,7 +215,43 @@ std::string TimeString() {
...
+  double newSize = (double)sizeInBytes / (double)den;
+  std::ostringstream oss;
+  oss.setf( std::ios::fixed, std::ios::floatfield );
+  oss.precision(3);
+  oss << newSize;
+  return oss.str() + std::string( BytePrefix[idx] );
+}
+
+// -----------------------------------------------------------------------------
+#ifdef __APPLE__
```
hainm: ha ha

Issue #177, opened 2015-11-16
Data that pytraj currently doesn't have access to but would like to


hainm:
related to https://github.com/Amber-MD/cpptraj/issues/175, https://github.com/Amber-MD/cpptraj/issues/93

I will keep updating checklist in 1st post and post the detail in following post.

- [ ] Action_NativeContact

drroe:
#72 seems like it belongs under this umbrella.

hainm:
### Native contact

- info about contacts

```bash
	Setup 33 native contacts:
		Atom ':1@CA' to ':2@CA'
		Atom ':1@CA' to ':11@CA'
		Atom ':1@CA' to ':12@CA'
		Atom ':2@CA' to ':3@CA'
...
```

- summary

```bash
#                     Contact  Nframes    Frac.      Avg    Stdev
       1          :1@OG_:1@HG      101        1     0.96 0.000365
       2        :2@NE1_:2@HE1      101        1     1.01 0.000366
       3           :1@N_:1@H2      101        1     1.01  0.00043
       4           :1@N_:1@H1      101        1     1.01 0.000364
       5           :1@N_:1@H3      101        1     1.01 0.000391
       6            :2@N_:2@H      101        1     1.01 0.000362
```

hainm:
#### Select dihedral, angle (bond too?) from Topology

The output should be ready to used for further calculation.

this output from cpptraj is good enough
```bash
         1:  2.000 0.00  1.0 :1@O      :1@C      :2@N      :2@H      (13,12,14,15) O -C -N -H 
E        2:  2.500 3.14  2.0 :1@O      :1@C      :2@N      :2@H      (13,12,14,15) O -C -N -H 
         3:  0.000 0.00  2.0 :1@C      :2@N      :2@CA     :2@HA     (12,14,16,17) C -N -CT-H1
        12:  0.000 0.00  2.0 :1@HA     :1@CA     :1@C      :2@N      (6,5,12,14) HP-CT-C -N 
```

(at least we could parse ":1@O      :1@C      :2@N      :2@H", ...)

hainm:
### Check structure

```python
In [10]: traj = pt.load('tz2.stretched.pdb')

In [11]: pt.common_actions.check_structure(traj)
1	 Warning: Atoms 222:NME_13@HH32 and 223:NME_13@HH33 are close (0.61)
```

The warning does not work with Jupyter notebook.

hainm:
### Action.Help()

[...]

PR #183, opened 2015-11-20
save reference frame numbers to Dataset for Clustering

hainm:
Related to #177

I am trying to do easy part.

```python
In [3]: pt.cluster.kmeans(traj, n_clusters=5, dtype='dataset')
Out[3]:
<pytraj.DatasetList with 2 datasets>
refframe
[4 0 8 6 7]

cluster_number
[1 1 0 0 0 0 3 4 2 2]
```


src/Analysis_Clustering.cpp
```diff
@@ -228,6 +229,10 @@ Analysis::RetType Analysis_Clustering::Setup(ArgList& analyzeArgs, DataSetList*
   // Output option for cluster info
   suppressInfo_ = analyzeArgs.hasKey("noinfo");
 
+  // Dataset to store representative cluster frames
+  clustersRefFrame_ = (DataSet_integer*) datasetlist->AddSet(DataSet::INTEGER, "refframe", "refframe");
+  if (clustersRefFrame_==0) return Analysis::ERR;
+
   // Dataset to store cluster number v time
```
drroe: This should come after the cluster num vs time data set is allocated and then use that DataSet name. All DataSets generated by an Action or an Analysis should have the same name. In this case you want to use the name of the `cnumvtime_` DataSet and add the Aspect 'refframe', e.g.
```C++
clustersRefFrame_ = datasetlist->AddSet(DataSet::Integer, MetaData(cnumvtime_->Meta().Name(), "refframe"));
```
Also, you don't need to make `clustersRefFrame_` a pointer to DataSet_integer. Just make it a `DataSet*`. Then you avoid having the `DataSet_integer.h` include in `Analysis_Clustering.h` and make things a bit more general. You can always cast to DataSet_integer inside `Analysis_Clustering.cpp` if need be.

hainm: Can you make another PR for this? I will close this one rather pushing new commit. It is easier for both of us. :d

Hai

hainm: Ps: i will keep in mind about aspect and casting dataset.

drroe:
> Can you make another PR for this? I will close this one rather pushing new commit. It is easier for both of us. :d

I'm delaying on this a little bit because I'm wondering if all clustering data shouldn't be stored in either one specialized DataSet or in multiple DataSets with different aspects. Former would allow cluster output to be used with the DataFile framework more easily, latter would be easier to implement in the short run but maybe harder to maintain. Maybe this should be a separate issue? Not sure...

PR #191, opened 2015-11-30
[WIP] Support Windows via mingw

swails:
This is far from complete so far, but it does iron out a few issues.

- [ ] Make readline optional and make it easy to turn readline build off in Makefile and Makefile_at (eventually needed for when configure supports MinGW compilers natively)

### DIFF
--- a/src/Action_RandomizeIons.cpp
+++ b/src/Action_RandomizeIons.cpp
@@ -1,7 +1,11 @@
 #include <cmath> // sqrt
-#include <cstdlib> // random, srandom
+#include <cstdlib> // srandom, random
 #include "Action_RandomizeIons.h"
 #include "CpptrajStdio.h"
+#ifdef __WIN32
+#   define srandom srand
+#   define random  rand
+#endif
 
 // CONSTRUCTOR
 Action_RandomizeIons::Action_RandomizeIons() :
--- a/src/FileName.cpp
+++ b/src/FileName.cpp
@@ -1,8 +1,11 @@
-#include <wordexp.h>
+#ifndef __WIN32
+#   include <wordexp.h>
+#endif
 #include <cstdio> // FILE, fopen
 #include "FileName.h"
 #include "CpptrajStdio.h"
 
+#ifndef __WIN32
 static void WexpErr(int err) {
   switch ( err ) {
     case WRDE_BADCHAR :
@@ -20,6 +23,7 @@ static void WexpErr(int err) {
       break;
   }
 }
+#endif /* __WIN32 */
 
 // COPY CONSTRUCTOR
 FileName::FileName( const FileName& rhs ) : fullPathName_(rhs.fullPathName_),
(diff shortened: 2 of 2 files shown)


hainm:
wow, +10000000000

swails:
This is a PR I'm hoping to collaborate on.  I expect this will need some collaboration.  It took me awhile to get MinGW set up on my Linux box (so I can compile Windows-native code on my Linux box), but it's working now.  I can compile a program with mingw-gcc that will only run under Wine on my machine.

Current stumbling block is that the bundled readline does not work with mingw.  So we need to make support for that optional.

Also I think that the changes that I made to `FileName.cpp` basically wipe out any globbing or environment variable expansion so that those won't work (on Windows only, obviously).

@drroe -- let me know if you need me to explain the workflow for collaborating on this PR.  The simplified description is that you will need to submit PRs to the `windows` branch on *my* fork.  There are other ways to do this, but the way I described is how I've done it in the past.

hainm:
@swails can you update things needed to be done in your 1st post? (bullet points)

src/Action_RandomizeIons.cpp
```diff
@@ -1,7 +1,11 @@
 #include <cmath> // sqrt
-#include <cstdlib> // random, srandom
+#include <cstdlib> // srandom, random
```
drroe: I wonder if it would be better to just use the `Random_Number` class (from `Random.h`)? The only reason `rand()` is used was to match ptraj results - not sure how critical `rand()` is to maintaining backwards compatibility since I don't think you could specify the seed in ptraj anyway. Would require some rewriting and testing so maybe it should be addressed in a new issue.

swails: +1 for reworking to use `Random_Number`.  For now it works so we can do this in a separate PR if you want to wait on it.  If you wanted to make this change quickly, then let's do it in this PR so we avoid merge conflict issues.

rmcgibbo:
(@swails for the curious, which mingw distro/fork did you end up using?)

swails:
My OS is Gentoo.  Gentoo's package manager (portage) compiles all packages (based on the port system for BSD, I believe).  They have a crossdev package that basically lets you run emerge with the mingw compilers.  This lets you construct an environment that you can chroot to where all of the libraries are Windows-compatible (of course you can do this with other distros where you just apt-get the mingw binaries, but Gentoo lets you do it through their package manager).  This lets you set up a chrooted environment where everything is built for Windows if you want to, but I only have the compilers right now.

I'm not sure what information you're looking for as far as "distro/fork" goes, but it's the standard Gentoo crossdev package x86_64-w64-mingw32 install (I can't tell what 64/32-bit compatibility the resulting binaries have).  The GCC version is 4.7 -- I tried 4.8 and 4.9 and neither would build successfully.

swails:
@drroe -- this wasn't meant to be merged yet (hence the WIP label).  My intention was to keep this PR open until the windows support was, at the very least, working.  I'll open up another PR when I get a chance to fix up the readline support, but don't merge it if it still says [WIP] in the title :)

rmcgibbo:
Got it -- thanks @swails.

drroe:
> (hence the WIP label)

@swails Woops - I see that now. On autopilot today...

swails:
I hear ya.  I'll label it in the PR text next time.  I don't think there's any harm in merging this, although if you could check the implications of my FileName changes that would be great (i.e., will my changes even work if I don't depend on env/~-expansion and globbing?)

PR #192, opened 2015-12-01
More work towards Windows compatibility

swails:
Ugly hack. But cpptraj now compiles with mingw (via -static) and runs
successfully under wine!

The readline-free support needs to be fixed in the Makefile and it needs to be
more thoroughly tested to make sure it behaves correctly.

But cpptraj now builds for Windows successfully as far as I can tell.

I had to do a lot of the steps by hand since we need better handling/automation for configure (and, by the way, -static is required to get this to run; otherwise we'll essentially need to require users to install mingw on their windows box)...

Who'd have thought that developing for Windows would be exciting? (Admittedly, this is developing on Unix for deployment on Windows -- perhaps it's just that I can develop for Windows without using it :)).

- [x] Still need better support for "automatic" building
- [x] Needs code review (and likely improvement -- the changes to the ReadLine interface were not as clean as I would have liked, although it does *work*)
- [x] Need to figure out math, zlib, and NetCDF library support in this environment
- [x] Need to add `-noreadline` option to configure
- [x] Need to add `-mingw` option to configure (or maybe `-windows`) which automatically turns on `-noreadline` and `-static` (among any other options we find to be necessary).
- [x] Need to figure out how to get testing infrastructure working in this environment

Don't merge yet.


rmcgibbo:
Have you considered trying this in a proper windows environment with Appveyor -- e.g. with msys2?

swails:
> Have you considered trying this in a proper windows environment with Appveyor -- e.g. with msys2?

I have a proper windows environment through VMs on my Linux Box, which is where I'll try (I also have access to a real-life Windows laptop).  Going through wine right now is my version of baby steps for this project.

Also, Appveyor has seemed way too slow for every other project I've seen it used for (MDTraj and OMM) for me to be able to use it for my kind of development (i.e., fumble around trying new stuff and compile/run to see if it works incrementally).  Once I have something more stable I'll try it in "real windows".  But if you have advice, I'm more than happy to try other stuff.

rmcgibbo:
Okay, that works then. Yes, Appveyor is quite slow.

I suspect you'll find though that it won't come out properly statically linked for things like gfortran/gmp/gmpfr, whatever the glibc is, and any of the dependencies like libz or libbz2. For ambermini, I had good luck using Carl K's mingw-w64 static toolchain (https://bitbucket.org/carlkl/mingw-w64-for-python/downloads), but overall most of my experiences with mingw have been pretty negative.

Obviously reconfiguring the build system to use msvc is a bit more of a headache.

hainm:
> But cpptraj now builds for Windows successfully as far as I can tell.

@swails can you post exact command(s) so I can try too.

swails:
I used this config.h file:

```
# config.h for cpptraj
# configured using: ./configure -nonetcdf -nomathlib -noarpack -nobzlib -nozlib gnu

CPPTRAJHOME=/home/swails/src/cpptraj
CPPTRAJBIN=/home/swails/src/cpptraj/bin
CPPTRAJLIB=/home/swails/src/cpptraj/lib

DBGFLAGS=
CC=x86_64-w64-mingw32-gcc
CXX=x86_64-w64-mingw32-g++
FC=x86_64-w64-mingw32-gfortran
CFLAGS= -O3 -Wall -DNO_READLINE -DNO_MATHLIB -DNO_ARPACK -DNO_ARPACK -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64  $(DBGFLAGS)
CXXFLAGS= -O3 -Wall -DNO_READLINE -DNO_MATHLIB -DNO_ARPACK -DNO_ARPACK -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64  $(DBGFLAGS)
FFLAGS=-ffree-form $(DBGFLAGS)

READLINE=noreadline
READLINE_HOME=readline

LIBCPPTRAJ=nolibcpptraj

FFT_DEPEND=pub_fft.o
FFT_LIB=pub_fft.o

LDFLAGS=         -lgfortran -w
SFX=.exe
```

Obviously change the name of the compilers to what you have.  This will also fail on the linking stage, since `noreadline` can't be found. So copy-and-paste that last linker command to the command-line and delete `noreadline`.  That should build cpptraj.exe.  However, I plan on having this fixed shortly so you won't have to do this.

cpptraj doesn't *really* respect the compiler variables, since it goes ahead and changes them to the canonical versions.  Which means you can't do `./configure CC=gcc44 CXX=g++44` and expect that to work.  I'll have to fix that up as well as create separate variables for the readline target and readline library to avoid the problem I just described.

hainm:
Thanks.

Hai

swails:
@hainm -- I've overhauled the configure script and Makefile so that now when I run

```
./configure CC=x86_64-w64-mingw32-gcc CXX=x86_64-w64-mingw32-g++ FC=x86_64-w64-mingw32-gfortran -windows -nobzlib -nozlib -nonetcdf -nomathlib -noarpack gnu
```

followed by `make install` it just works.

hainm:
cool. I will try.

swails:
@drroe -- I made a number of changes to how `configure` reads and uses existing environment variables, and I think it's more consistent this way compared to how other programs use them:

1. Existing environment variables `CC`, `CXX`, and `FC` *always* take precedence and are never altered, even if you supply compilers on the command-line.  If present, the `compiler` option will set the requisite compiler flags based on the desired compilers.

2. Existing CFLAGS, FFLAGS, CXXFLAGS, and LDFLAGS variables are *added* to what `configure` sets by default (and added to the end so that they will always override what cpptraj wants to put if they contradict each other).  The previous behavior was that, if present, these env variables would replace the standard cpptraj flags.

\# 1 is needed for the command I gave @hainm because decorated compiler names won't be recognized given the current logic, and this is the only way that we can make sure we get the correct compiler flags *and* the desired compilers.

For \# 2, the most important bit there is LDFLAGS, which the user will almost always want to *supplement*, not replace.  Although I think this is more or less true for each of the variables.

swails:
OK, this is in a good enough state to merge, and I'm working on getting this cross-compiling working on AmberTools in general.  I'll clean up the rest of the stuff with the libraries there.

swails:
actually I take that back -- I have NetCDF working now and I want to get cpptraj building with that lib (it doesn't currently)

rmcgibbo:
@swails: I logged onto windows and got a native binary working from this PR using msys2 using the following steps:

Download and unzip the netcdf-4.3.3 source distribution from http://www.unidata.ucar.edu/downloads/netcdf/index.jsp, and then do everything in the msys2 shell (launched via `C:\msys64\mingw64_shell`)

```
cd /path/to/netcdf/
 ./configure --enable-static --disable-netcdf-4 --prefix=/mingw64
make install
```

```
pacman -S mingw-w64-x86_64-gcc-fortran mingw-w64-x86_64-openblas mingw-w64-x86_64-arpack
```

```
cd /path/to/cpptraj
./configure -windows --with-netcdf=/usr/local
```
Then I needed to manually modify `LDFLAGS` in `config.h` to
```
LDFLAGS= -larpack -lnetcdf -lopenblas -lbz2 -lz  -lgfortran -lquadmath -w  -static
```

before running `make`.

rmcgibbo:
Also, readline works fine using `pacman -S mingw-w64-x86_64-ncurses mingw-w64-x86_64-readline` and then adding `-lreadline -lncurses` to the link line and `-I/mingw64/include/readline/` to the clfags.

swails:
Thanks for the info!  Maybe I should be putting my efforts into msys2 with Amber rather than cross-compiling with mingw on Linux.

I did make some more progress with the cross-compiling, though.  I got NetCDF 4.1.3 installed (this was before they split the Fortran bindings out to another package).  I could get the latest C bindings to install correctly, but not the Fortran bindings (which I need for sander).  I also got zlib and bz2 installed with mingw, and I'm trying to see if they're working now.

I have access to a Windows laptop I can try out next week, and I'll give msys2 a try then.  It would be nice to be able to get some Windows-compatible conda packages for more Amber components (possibly allowing us to do away with the ambermini package).

rmcgibbo:
>  It would be nice to be able to get some Windows-compatible conda packages for more Amber components (possibly allowing us to do away with the ambermini package).

Yeah, absolutely. I concur.

rmcgibbo:
One of the implications is with the readline changes though. Ideally it would be nice to add config options to support linking against an externally-supplied `libreadline`, instead of always either building the bundled copy or simply turning it off.

Another is that with `openblas`, both blas and lapack are provided in a single library, and that library is named `libopenblas.{so/a/dll/lib}`, not `libblas` or `liblapack`, so that requires a different link line that what the config script currently makes.

swails:
It's also available in msys2 through pacman, which is probably the easier way to build cpptraj.

hainm:
@swails do you intend to add `--with-openblas=` too?

swails:
No.  You can use `--with-blas` or `--with-lapack`.  Both are valid, both apply, and both work.

hainm:
can you post an working command so I can try on Linux?

This does not work for me:
```bash
$ ./configure --with-netcdf=~/anaconda3/ --with-blas=~/anaconda3 --with-lapack -openblas gnu
```

```
  Please check your compiler settings or configure flags. Error follows:
  --------------------
/usr/bin/ld: cannot find -larpack
collect2: ld returned 1 exit status
```

hainm:
just retry my stupid mistake, sounds good for me. thanks.

`./configure --with-netcdf=~/anaconda3/ --with-blas=~/anaconda3 --with-lapack -openblas -noarpack gnu`

swails:
@rmcgibbo -- I was able to modify configure so that it works for Windows like this:

```
$ ./configure --with-netcdf=/usr/local -openblas -windows --with-readline=/mingw64/usr gnu
```

And everything built fine (including with readline).  But when I tried to actually *run* cpptraj in the msys2 terminal, it just sat there.  Same thing with `python` -- it seems to choke on interactive CLIs as far as I can tell (and I don't know why).  But I've heard this is a known problem.  It also seems like using readline for cpptraj breaks cpptraj in cmd.exe (it worked fine for me without readline using the cross-compiled binary).  It just exits without taking any input...

In any case, I'll try updating the appveyor YAML file to use configure tomorrow.  Thanks for your help!

rmcgibbo:
Can you try downloading the binary from the `artifacts` section on Appveor built there https://ci.appveyor.com/project/drroe/cpptraj-aof9y/build/1.0.4/artifacts? Does it have the same problem?

rmcgibbo:
(Also, should the `-windows` flag be automatically detected if you're actually on windows?)

rmcgibbo:
@swails I see something similar, but I'm not sure what's going on. When I cpptraj run from inside the msys2/mingw bash shell, it just hangs. When I run it from powershell, I see the logo and date, and then it just exits.
```
PS C:\Users\Administrator\cpptraj> .\bin\cpptraj.exe --interactive

CPPTRAJ: Trajectory Analysis. V16.00b
    ___  ___  ___  ___
     | \/ | \/ | \/ |
    _|_/\_|_/\_|_/\_|_

| Date/time: 12/07/15 23:14:19

        Loading previous history from log 'cpptraj.log'
> --------------------------------------------------------------------------------
To cite CPPTRAJ use:
Daniel R. Roe and Thomas E. Cheatham, III, "PTRAJ and CPPTRAJ: Software for
  Processing and Analysis of Molecular Dynamics Trajectory Data". J. Chem.
  Theory Comput., 2013, 9 (7), pp 3084-3095.

PS C:\Users\Administrator\cpptraj>
```

rmcgibbo:
I think the readline thing is semi-fixed with termcap, in that the interactive mode works in the windows cmd shell and powershell, although it doesn't for me in the msys2 shell (or in wine). Here's another issue: 

I [added netcdf](https://github.com/Alexpux/MINGW-packages/tree/master/mingw-w64-netcdf) to the msys2 pacman, and I figured that to be helpful I'd compile it with as many features as possible, including netcdf-4 (hdf5). Perhaps this was a mistake. With shared libraries, it works great, but with static libraries you have a much trickier time coming up with the proper link line (the current test in `configure` isn't actually sufficient). In addition to linking against `libnetcdf.a`, you have to explicitly link all of its dependencies. And you need to do it in the proper order. What the dependencies are depends on how netcdf was built, and if it was built against hdf5, how that was built. As far as I can tell, the fact that cpptraj doesn't use hdf5 doesn't actually matter, because you still get unresolved external symbols. netcdf comes with a tool that's supposed to handle this and tell you the proper link line, called `nc-config`, but it's broken for static libraries and doesn't actually give you the right answer.

For the pacman-installable netcdf, the link line you need for static linking is actually `-lnetcdf -lhdf5_hl -lhdf5 -lszip`.

This isn't really a windows specific issue -- the same issue will come up if you want to make a truly statically linked executable on linux.

rmcgibbo:
I guess the windows way is probably to dynamically link the dependencies and then package up the exe and its libraries into a bundle with an GUI installer.

swails:
@rmcgibbo -- that is exactly what I was seeing.  I will try with termcap and update the PR with that option.  I still think the msys2 shell is just broken for any program reading from stdin.  Not sure why (same thing happens if you try to run Anaconda's Python from msys2, but it works fine under powershell, assuming that's the name given to cmd.exe).

> I added netcdf to the msys2 pacman, and I figured that to be helpful I'd compile it with as many features as possible, including netcdf-4 (hdf5). Perhaps this was a mistake. With shared libraries, it works great, but with static libraries you have a much trickier time coming up with the proper link line (the current test in configure isn't actually sufficient). In addition to linking against libnetcdf.a, you have to explicitly link all of its dependencies. And you need to do it in the proper order. What the dependencies are depends on how netcdf was built, and if it was built against hdf5, how that was built. As far as I can tell, the fact that cpptraj doesn't use hdf5 doesn't actually matter, because you still get unresolved external symbols. netcdf comes with a tool that's supposed to handle this and tell you the proper link line, called nc-config, but it's broken for static libraries and doesn't actually give you the right answer.
> 
> For the pacman-installable netcdf, the link line you need for static linking is actually -lnetcdf -lhdf5_hl -lhdf5 -lszip.
> 
> This isn't really a windows specific issue -- the same issue will come up if you want to make a truly statically linked executable on linux.

I went through this exact same thing a year or two ago when we updated the Amber-bundled NetCDF to version 4.  What I've been doing is using a version of NetCDF 4.1.3 that I compiled on my Linux machine (with the `x86_64-w64-mingw32` host) without NetCDF-4 and HDF5.  The reason I used that older version is because I simply couldn't get the latest version of the Fortran bindings working on my Linux machine, and 4.1.3 did everything I needed it to.

The reason I want the Fortran bindings is for the rest of Amber (sander and pmemd, specifically).  And all I have to do is package up the zlib, bzlib, and NetCDF libraries I built for Windows and deploy it wherever I decide to build Amber, and it works (I've managed to get all of Amber except nab ported).

swails:
> (Also, should the -windows flag be automatically detected if you're actually on windows? Maybe it should be -mingw anyways or something -- people might think it's visual studio and expect configure to produce VS solution files or an NMakefile)

Fair points.  I would like `-windows` to be auto-detected on Windows, but I didn't know a reliable way to set that flag.  I can't use `python`, since I can't assume that exists.  I also can't use uname to my knowledge, since I don't know what output to expect from whatever shell is being used...

I didn't really care *that* much since I figured anyone interested in running on windows would want pre-built binaries...

configure
```diff
@@ -583,6 +618,12 @@ while [[ ! -z $1 ]] ; do
 #      HDF5LIB="-I$VALUE/include $VALUE/lib/libhdf5_hl.a $VALUE/lib/libhdf5.a -lm"
 #      echo "Using HDF5LIB: $HDF5LIB"
 #      ;;
+    "--with-readline")
```
rmcgibbo: Do you recall the reason for bundling readline in the first place? Surely it's one of the most widely deployed packages in any linux -- isn't it like a dependency of the shell?

rmcgibbo:
Also, `-static` is not actually *required* with mingw any more than it is with linux/gcc. As on linux, it's got the same tradeoffs (harder to know the proper link line and generally trickier to compile with static, but easier to redistribute binaries). It might be a good default tho.

configure
swails: > Surely it's one of the most widely deployed packages in any linux -- isn't it like a dependency of the shell?

@drroe will have to weigh in here.  I know that Amber was already bundling and shipping a copy of readline as part of the deprecated and now-removed sleap program (a tleap rewrite that *was* better in many ways, except it was buggy and nobody used it so it never got debugged).  It may have just been simple to copy that over and add it to cpptraj.

Also, just because numerous tools had readline support doesn't mean the dev headers and libraries were installed on each system...  It's such a lightweight dependency that we probably thought "what's the harm?".  But Dan would have the final say on why it was done.

swails:
It *is* needed to cross-compile on my Linux machine for Windows, though.  Which is why it worked its way in there :).  Now that cross-compilation is determined more reliably, we can turn it on only in that instance.

The problem with the default as-is is that you need to hack the config.h file to remove it (there's no configure option to turn it off as of yet).

configure
rmcgibbo: Amber(Tools) as a distribution builds a lot of dependencies (of relevance to cpptraj: blas, lapack, arpack, netcdf, etc) that it doesn't depend on the system for, but readline here is slightly different  in that a copy is embedded inside of the cpptraj source tree, unlike those others.

I checked Debian, and there readline is not a dependency of bash. On Arch, it is though. Either way, a lot of the system depends on it though, from python to whatever.

mpharrigan: The dev headers generally still aren't installed (?)

hainm: yes, even with python.

swails: > but readline here is slightly different in that a copy is embedded inside of the cpptraj source tree, unlike those others.

The older copy of readline in the git tree is in `AmberTools/src/gleap/freelib/src`.  Since that isn't distributed anymore, it was copied to the cpptraj source tree.

rmcgibbo: Fair enough about the headers. I think cpptraj only uses 2 or 3 functions from readline, so you could just extern declare their prototypes even without having the headers. This is common for BLAS.

drroe: > Do you recall the reason for bundling readline in the first place? Surely it's one of the most widely deployed packages in any linux -- isn't it like a dependency of the shell?

@rmcgibbo So the `readline` that ships with Amber (for cpptraj, formerly for sleap/gleap as @swails mentioned) is actually a stripped-down version of readline edited by Wei Zhang (I think) with only the core functionality there. I briefly looked into linking in `readline` externally but it seemed like more work than it was worth at the time. I'm not opposed to changes, but using this `readline` has been the path of least resistance thus far :-)

rmcgibbo: Anyways, this is not a big deal. Bundling it probably eliminates more headaches than it causes for most configurations. Windows is the exception here because the embedded copy does not compile under Mingw, but the Msys2 system helpfully includes a copy with the necessary platform-specific patches.

drroe:
Sorry, closed this accidentally.

hainm:
@swails can  you make another PR to make openblas work so Dan can merge to master?

I want to build pytraj with cpptraj and openblas (already in conda) to avoid building blas, lapack. For some reasons I messed up with gfortran library with blas and lapack in casegroup, so I need to build pytraj in another machine (or from travis). But the library in casegroup is quite outdated (I spent too much time on this issue), so if build pytraj on travis and then install to casegroup, I got

`ImportError: /usr/lib64/libgfortran.so.3: version `GFORTRAN_1.4' not found`.

So I just try to build pytraj on old machine and this will work with newer one. :D

rmcgibbo:
@hainm feel free to file a pr against omnia if you have want to build binaries on CentOS 5.

hainm:
thanks Robert.

swails:
I think we can go ahead and merge this.  We don't have that much more time before the Amber code freeze.

rmcgibbo:
Do you want to modify the Appveyor.yml file to remove the custom config.h file that's written, and confirm that this generates a proper config.h in that environment?

rmcgibbo:
Also, at that point, it might make sense to just delete https://github.com/Amber-MD/cpptraj/tree/master/Windows -- given that `git` says that subdirectory was last touched 4 years ago, I suspect it's bitrotted anways.

hainm:
> Also, at that point, it might make sense to just delete https://github.com/Amber-MD/cpptraj/tree/master/Windows -- given that git says that subdirectory was last touched 4 years ago, I suspect it's bitrotted anways.

agree.

drroe:
> Also, at that point, it might make sense to just delete https://github.com/Amber-MD/cpptraj/tree/master/Windows 

This makes sense to me. @swails if you agree could you do that as part of this PR? Thanks.

swails:
> This makes sense to me. @swails if you agree could you do that as part of this PR? Thanks.

Done.  I also merged master and rebased, so this should be a clean FF merge

dacase:
Above info is all very impressive, but daunting to a newcomer.  There are several suggestions for how to handle netcdf, a couple for readline, blas/openblas/arpack, etc.  This issue stream also has a mixture of recommendation for cross-compiling on Linux as well as for using the msys2 shell on windows.  Is some summary possible, hopefully using features that have been put into the configure2 script and pushed to the git master at SDSC?

You post here or at http://ambermd.org/pmwiki/pmwiki.php/Main/Windows.

Thanks!  and kudos for all the hard work...I just don't want to have all this get lost.

PR #193, opened 2015-12-01
Randomize ions

drroe:
This PR makes the `randomizeions` action use the `Random_Number` class for random number generation. Also cleans up `Action_RandomizeIons` a bit.


swails:
Sweet!

PR #197, opened 2015-12-05
Build and test cpptraj on Windows with Appveyor

rmcgibbo:
This some of the ideas from https://github.com/Amber-MD/cpptraj/pull/192 to build on Windows under Appveyor-CI. It just skips running the configure script for now because I figured that was still in flux with #192.

Here's the build log: https://ci.appveyor.com/project/rmcgibbo/cpptraj/build/1.0.88


swails:
@drroe -- you need to go to appveyor.com and sign up for a free open source account and then activate it for this Github project.  It's a lot slower than Travis, but still finished in <9 minutes.  It's a nice CI check for Windows compatibility.

+1 on this -- looks good to merge to me.

rmcgibbo:
I think the remaining todos are
 - figure out what changes should be made to the `configure` script to better support this toolchain (readline, blas, etc)
 - change the `appveyor.yml` to use `configure` instead of hardcoding `config.h`.
 - figure out how to not require libnetcdf to be rebuilt every time. The cleanest solution would be to just submit a PR with the build recipe to the repo that manages the msys2 packages installable through `pacman`.

hainm:
I just have very naive question: Can we just use things (netcdf, mingw, ...) in Anaconda for window?

drroe:
@rmcgibbo Thanks for this!

rmcgibbo:
@drroe i think you still might need to sign up for appveyor or [activate](http://www.appveyor.com/docs) the project for this to work complete. Ideally you'll see two checks performed on each PR, one by the travis service, and one by appveyor.

rmcgibbo:
@hainm There is no inherent incompatibility with using mingw and msys2 together with Anaconda.

The golden rule on windows is that you shouldn't, in general, mix compiler versions when making libraries that are linked together. This applies to either compiling some files using one compiler and some files using a different compiler and then linking them together to make a single executable, or to dynamically loading a plugin (e.g. a python extension module) that was compiled using one compiler from a program built using a different compiler. For Python, this means that you need to, in general, extensions need to be build with the same version of the compiler that was used to build Python (VS2008 for py27, VS2010 for py3{3-4}, VS2015 for py35).

If you're just going to build a single standalone executable (not a shared library), these compatibility requirements don't matter.

Anaconda is a distribution that _mostly_ contains python extensions, so the compatibility issues apply to them. But there's no problem with making conda packages for standalone executables, so in that case the compiler doesn't matter.

Msys2 is a cygwin-derived environment for building Windows-native software with the mingw-w64 compiler, so it comes in quite handy here. Its package manager also has pre-compiled packages (built with mingw-w64) for lots of dependencies. Most of them seem to be compiled as static libraries, so that's excellent because it means that you can compile executables that don't have any external dependencies (and the user doesn't need to have msys2 installed).

hainm:
> Msys2 is a cygwin-derived environment for building Windows-native software with the mingw-w64 compiler, so it comes in quite handy here. Its package manager also has pre-compiled packages (built with mingw-w64) for lots of dependencies. Most of them seem to be compiled as static libraries, so that's excellent because it means that you can compile executables that don't have any external dependencies (and the user doesn't need to have msys2 installed).

great. thanks for the info.

drroe:
> i think you still might need to sign up for appveyor or activate the project for this to work complete. Ideally you'll see two checks performed on each PR, one by the travis service, and one by appveyor.

@rmcgibbo Yeah, I thought I had done this over the weekend but turns out I was just checking my local cpptraj fork (oops). Set it up this morning and it appears to be working correctly now.

PR #199, opened 2015-12-07
fixed grid origin inconsistency

gosldorf:
Removed '-0.5*gridspacn_' from the defining origin lines, since it would un-intuitively offset the grid origin.


drroe:
@gosldorf I just noticed something about the cpptraj tests - currently Travis-CI is only running the basic test cases. There are a few tests that are not included, including GIST (this is for weird historical reasons and will be changed soon). For now can you do the following in your `gist_patch` branch for this commit?
```
cd $CPPTRAJHOME/test
make clean test.all
```
That will run a few additional tests, including the GIST test. Thanks!

drroe:
Thanks!

PR #201, opened 2015-12-07
Switch ncurses for termcap

rmcgibbo:
@swails this seems to fix the interactive issue on windows for me. (based on http://sourceforge.net/p/msys2/tickets/32/, I think ncurses is broken in msys2)


swails:
Cool.  I'll give it a try.  Thanks!

PR #202, opened 2015-12-08
Do not depend in isatty on Windows

rmcgibbo:
This fixes https://github.com/Amber-MD/cpptraj/pull/192#issuecomment-162701749. It seems that you cannot depend on the output of `isatty` -- it works fine in the regular windows shell, but not in the shell emulators included with msys2 or cygwin, because they have to use pipes to emulate the terminal protocol. There is some discussion of this here: https://code.google.com/p/mintty/issues/detail?id=56. Python includes a "-i" flag that forces interactive mode even if stdin does not appear to be a tty for this purpose. With this, it seems to work fine in either of the shells or shell emulators.

I don't know the exact logic you want here with interactive vs. piping from stdin, but essentially, on windows, if user passes `--interactive`, it should be honored. Even if stdin does not "appear" to be a tty.

### DIFF
--- a/src/Cpptraj.cpp
+++ b/src/Cpptraj.cpp
@@ -315,15 +315,14 @@ Cpptraj::Mode Cpptraj::ProcessCmdLineArgs(int argc, char** argv) {
   if (interactive) {
     // User explicitly requested ``--interactive``. Do not check isatty.
     return INTERACTIVE;
-  }
-  if (!hasInput && isatty(fileno(stdin))) {
-     // No input and stdin is a console tty
-    return INTERACTIVE;
-  } else {
-    // "" means read from STDIN
-    Command::RetType c_err = Command::ProcessInput( State_, "" );
-    if (c_err == Command::C_ERR && State_.ExitOnError()) return ERROR;
-    if (c_err == Command::C_QUIT) return QUIT;
+  } else if (!hasInput) {
+    if (isatty(fileno(stdin)))
+      return INTERACTIVE;
+    else {
+      Cmd::RetType c_err = Command::ProcessInput( State_, "" );
+      if (c_err == Cmd::ERR && State_.ExitOnError()) return ERROR;
+      if (c_err == Cmd::QUIT) return QUIT;
+    }
   }
   return BATCH;
 }
--- a/src/Cpptraj.cpp
+++ b/src/Cpptraj.cpp
@@ -311,23 +311,19 @@ Cpptraj::Mode Cpptraj::ProcessCmdLineArgs(int argc, char** argv) {
       if (c_err == Command::C_QUIT) return QUIT;
     }
   }
-  // Determine whether to enter interactive mode
-  if (!hasInput || interactive) {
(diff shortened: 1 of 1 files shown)


drroe:
Cool, thanks for the fix.

> I don't know the exact logic you want here with interactive vs. piping from stdin, but essentially, on windows, if user passes --interactive, it should be honored. Even if stdin does not "appear" to be a tty.

The desired behavior is that interactive mode is entered if there has been no "actionable input" (i.e. actions, analyses, output trajectories, etc). However, if input will be coming from something that is not a terminal (like a file via `< file`) assume that contains the actionable input and use batch mode instead. Not sure if this is the ideal behavior, but at the time it was implemented it seemed like a good guess for what the user wanted (and pretty closely reproduced how ptraj would handle similar situations, even though the interactive mode in ptraj wasn't very developed).

drroe:
(Based on the WIP label I'm assuming you don't want a merge yet).

rmcgibbo:
I wasn't sure how you'd prefer to guess between interactive and assuming that it's a pipe from `stdin`. This PR would work, but a different option would be like.

```
if ('--interactive flag' is given explicitly)
  return interactive
else
   use the existing code
```

This would remove the need for the `ifdef _WIN32`, but change the behavior slightly.

rmcgibbo:
Okay, I changed the behavior slightly such that `--interactive` overrides checking `isatty` -- if the user passes `--interactive` you know that's what they're trying to do. I think this is slightly better.

swails:
Ah, this would explain both Python's and cpptraj's behavior in the msys2 terminal.

Travis seems to be failing because cpptraj is expecting interactive input, though.  Any idea what's going on here?

src/Cpptraj.cpp
```diff
@@ -311,17 +311,19 @@ Cpptraj::Mode Cpptraj::ProcessCmdLineArgs(int argc, char** argv) {
...
-      Command::RetType c_err = Command::ProcessInput( State_, "" ); 
-      if (c_err == Command::C_ERR && State_.ExitOnError()) return ERROR;
-      if (c_err == Command::C_QUIT) return QUIT;
-    }
+  // Determine whether to enter interactive mode.
+  if (interactive) {
+    // User explicitly requested ``--interactive``. Do not check isatty.
+    return INTERACTIVE;
+  }
+  if (!hasInput && isatty(fileno(stdin))) {
```
drroe: I think maybe this needs to be:
```C++
  if (!hasInput) {
    if (isatty(fileno(stdin)))
      return INTERACTIVE;
    else {
      Cmd::RetType c_err = Command::ProcessInput( State_, "" );
      if (c_err == Cmd::ERR && State_.ExitOnError()) return ERROR;
      if (c_err == Cmd::QUIT) return QUIT;
    }
  }
  return BATCH;
```
Otherwise whenever we are not entering batch mode because input is from a terminal, we're automatically trying to get input from a STDIN redirect, and therefore never getting down to BATCH right?

PR #203, opened 2015-12-09
Stronger check netcdf

rmcgibbo:
It turns out that the existing netcdf check doesn't actually detect a bunch of possible link time errors. For example, if you statically compile netcdf with netcdf-4 support, the link line isn't correct and you get unresolved hdf5 symbols. But for whatever reason, the check with only `nc_strerror` doesn't flag this.


drroe:
Cool, thanks!

PR #205, opened 2015-12-11
New command structure

drroe:
This PR is a major refactoring of how commands are handled in cpptraj. Previously, commands were stored in a static array in Command.cpp and made use of function pointers to call the necessary command or help function.

The new approach separates the concept of a "command" and the underlying implementation, called a DispatchObject. This takes into account the fact that not all commands want to do the same thing; Actions will usually be sent to a queue, while commands like `trajin` or `parmstrip` will be executed immediately, etc. Now commands are stored in the `Cmd` class, which holds all necessary information to execute the command. There are currently 4 types of commands: ACT (action), ANA (analysis), EXE (immediate execution), and DEP (deprecated).

The biggest benefit from a development standpoint is now it is much clearer what different types of commands require; it also ensures that the command is properly implemented at compile time. Also, the code for individual commands has now been split among separate files, which improves the overall readability of the code (instead of having everything in the monolithic `Command.cpp`). Another minor benefit is that in `Command.cpp` it is now easy to see all keywords associated with an underlying command type since they are on the same line, e.g.
```C++
Command::AddCmd( new Exec_AtomInfo(), Cmd::EXE, 3, "atoms", "atominfo", "printatoms" )
```
Aside from these benefits, this just feels like a "more C++" way to implement command logic.

In addition, the Analysis class has now been modified so that Setup() takes a class instead of having pointers passed in (similar to how Action::Init() was modified). This will make it easier to modify if other information ever needs to be passed to Analysis::Setup().

@hainm This is a major API change so I have incremented the internal version number, but I haven't tested with pytraj yet. If you would like to test this PR with pytraj before I merge I would appreciate it - thanks!


hainm:
> In addition, the Analysis class has now been modified so that Setup() takes a class instead of having pointers passed in (similar to how Action::Init() was modified). This will make it easier to modify if other information ever needs to be passed to Analysis::Setup().

this will break `pytraj` for sure. Let me update.

hainm:
And can you convert lyx file to 'html' (then update in your repo) so I can update here: http://amber-md.github.io/pytraj/latest/CpptrajDevelopmentGuide.html

lyx guide: http://wiki.lyx.org/Tools/ELyXer

hainm:
@drroe 

> typedef ActionInit AnalysisSetup

why not `typedef ActionInit AnalysisInit`

hainm:
I just update pytraj's code in another branch. Will wait for your new update so I can install cpptraj.

Hopefully you don't change your mind to massively break cpptraj's API several times next year.

drroe:
> Hopefully you don't change your mind to massively break cpptraj's API several times next year.

@hainm Believe me, I don't do this just for the sake of causing massive issues. Whenever I make an API change it's because I believe that any future developers (fingers crossed) will have an easier time with it. I don't want to end up with a LEaP situation - in my opinion awesome code but difficult to grok. The command API has until this point been a bit of a free for all hack - this PR formalizes commands a bit. Essentially, the development of pytraj (which I think is an outstanding project) has finally spurred me into making all the API changes to cpptraj that I thought would be good but never did because I was the only person working on the code. So in the short term it's a bit of a headache but my hope is that in the long term this is a net benefit.

hainm:
:100: very persuasive. thanks for detail explanation. 

Do you foresee that we can finalize cpptraj API before Amber16 release?

> The command API has until this point been a bit of a free for all hack - this PR formalizes commands a bit. 

pytraj barely use command API, so I am not worried about this. But the Action and Analysis API should be stable. 

> So in the short term it's a bit of a headache but my hope is that in the long term this is a net benefit.

agree.

drroe:
> Do you foresee that we can finalize cpptraj API before Amber16 release?

I don't forsee any changes to the Action/Analysis API in the future. In fact this PR in theory means any such changes will be in the form of additions to ActionSetup, so if stuff does need to change it will be less painful.

PR #210, opened 2015-12-17
Hbond indices

hainm:
This is my proposal to add atom indices to hbond.

Benefit:
- `pytraj` can use hbond atom indices to compute angle, distance

- viewing: for example, I want to colorize all atoms that form hbond, I can pass atom index array to different programs to do the coloring. (given that those programs do not understand amber mask syntax).

So just adding **hidden** keyword 'atomindex' to dump atom indices to DataSet_integer (pytraj can use numpy to reshape to (n_hbonds, 3)).

### DIFF
--- a/src/Action_Hbond.cpp
+++ b/src/Action_Hbond.cpp
@@ -40,6 +40,7 @@ void Action_Hbond::Help() const {
           "\t[solventdonor <sdmask>] [solventacceptor <samask>]\n"
           "\t[solvout <filename>] [bridgeout <filename>]\n"
           "\t[series [uuseries <filename>] [uvseries <filename>]]\n"
+          "\t[atomindex]\n"
           "  Hydrogen bond is defined as A-HD, where A is acceptor heavy atom, H is\n"
           "  hydrogen, D is donor heavy atom. Hydrogen bond is formed when\n"
           "  A to D distance < dcut and A-H-D angle > acut; if acut < 0 it is ignored.\n"
@@ -53,6 +54,8 @@ void Action_Hbond::Help() const {
           "  If donorhmask is specified atoms in that mask will be paired with atoms in\n"
           "  donormask instead of automatically searching for hydrogen atoms.\n");
 }
+
+// Add atomindex to get atom indices for pytraj (hide from cpptraj man)
 
 // Action_Hbond::Init()
 Action::RetType Action_Hbond::Init(ArgList& actionArgs, ActionInit& init, int debugIn)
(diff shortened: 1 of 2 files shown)


hainm:
anyway, just try to make a PR so you (Dan) can prioritize this. :D

src/Action_Hbond.cpp
```diff
@@ -76,6 +80,9 @@ Action::RetType Action_Hbond::Init(ArgList& actionArgs, ActionInit& init, int de
   useAtomNum_ = actionArgs.hasKey("printatomnum");
   acut_ = actionArgs.getKeyDouble("angle",135.0);
   noIntramol_ = actionArgs.hasKey("nointramol");
+  if (actionArgs.hasKey("atomindex")) {
+    AHDIndices_ = (DataSet_integer*) init.DSL().AddSet(DataSet::INTEGER, "atomindex");
```
drroe: @hainm The convention is that all DataSets generated by an Action should have the same name, so e.g. in `hbond` the number of solute-solute hbonds, number of solvent-solvent hbonds, and number of bridging interactions are given the same name with different aspects:
```C++
NumHbonds_ = init.DSL().AddSet(DataSet::INTEGER, MetaData(hbsetname_, "UU"));
NumSolvent_ = init.DSL().AddSet(DataSet::INTEGER, MetaData(hbsetname_, "UV"));
NumBridge_ = init.DSL().AddSet(DataSet::INTEGER, MetaData(hbsetname_, "Bridge"));
```
So what you should really do here is:
```C++
AHDIndices_ = (DataSet_integer*)init.DSL().AddSet(DataSet::INTEGER, MetaData(hbsetname_,"atomindex"));
```

drroe:
I'll try and pull this PR down today and test it out. Seems OK for now - eventually it might be better to make a custom data set for this but we can talk about that later.

src/Action_Hbond.cpp
hainm: @drroe 

I change to yours and got this

`Internal Error: DataSet has no name.
Error setting up data set [atomindex].
`

drroe: I forgot to mention this before; the reason you get this error is you are setting up your DataSet too early. The `hbsetname_` variable isn't assigned until after all keywords and masks are processed; this is also a convention (and for a wonder is actually mentioned in the dev. guide in the `Adding Actions` section). So, you should set up your DataSet after the comment `// Setup datasets`. Use a `bool` to save whether your keyword is there or not:
```C++
bool useAtomIndex = actionArgs.hasKey("atomindex");
```
then probably after the `if (calcSolvent_)` section, put
```C++
if (useAtomIndex)
  AHDIndices_ = (DataSet_integer*)init.DSL().AddSet(DataSet::INTEGER, MetaData(hbsetname_,"atomindex"));
```
Make sense?

hainm: thanks. just updated PR. The dataset has name 'HB00000[atomindex]' now.

hainm: but the order of the indices does not match to cpptraj's label for each hbond.

For example:

this is the indices if specifying `atomindex`
```python
[[ 50   3   0]
 [  9  14  13]
 [ 89  48  47]
 ..., 
 [167 212 211]
 [170 212 211]
 [170 213 211]]
```

vs the indices from parsing cpptraj's hbond output: 'SER_1@OG-TRP_2@N-H', 'GLU_5@O-THR_3@OG1-HG1', ...

```python
[9, 14, 13], [89, 48, 47], [50, 3, 0], [103, 112, 111], [47, 100, 99], [36, 52, 51], [167, 130, 127], [170, 212, 211], [170, 168, 167], [167, 212, 211], [170, 213, 211], [87, 91, 90], [110, 134, 133]]
```

hainm:
May be we need more though about this. So I close this PR.

PR #212, opened 2015-12-22
Fix GIST exception problems, add new GIST test.

drroe:
The GIST action Print() routine contained a bunch of try/catch blocks, presumably to avoid divide by zero errors. However, they didn't work as intended.

@gosldorf  Divide by zero is not an exception you can catch in standard C++ so the program would still crash out. All try/catch blocks in Print() have been removed and replaced with explicit checks for divide by zero. I'm pretty sure I got all the checks for zero correct but since you wrote the original code I would appreciate it if you would look it over and see if I missed anything. Thanks.

@cnnguyen I've also added the tetrahedral geometry GIST test that I believe you gave me a while back (this test is how I found the issue being fixed by this PR). I had to update the output for some of the recent changes to GIST; in particular g_H has gone from 0.0 to 15.015; seems reasonable but I'd appreciate if you could double check for me. Thanks!


gosldorf:
Ah. Sorry about that.

The intent of those catch lines is to determine whether a voxel is on the
boundary of the grid and if so skip entropy calculations that require
neighboring voxels (as there would be a reduced count of neighbor space
within the stored grid if we're on the edge). Now that I'm looking at it in
an if-then format I'm not sure why I used try-catch instead.

In any case I took a look at the changes and tested it out on a couple of
test systems and it looks good.

Also sorry about responding after the fact,

--Steve

drroe:
> Also sorry about responding after the fact,

@gosldorf Not a problem! Just wanted to make sure you knew about it since you wrote the original code. Thanks for double checking the changes!

PR #217, opened 2016-01-03
expr index in bash not portable to BSD, replace with awk

### LINKED ISSUE #216: ./configure error on OS X
I get a syntax error when running configure on OS X

```
$ ./configure --with-netcdf=sfd

Assuming 64 bit architecture.
expr: syntax error
Unrecognized OPTION: --with-netcdf=sfd
```

When I run with `bash -x -e`, I see that the error is at [line 400](https://github.com/Amber-MD/cpptraj/blob/master/configure#L400). It [looks like](https://discussions.apple.com/thread/923299?tstart=0) BSD `expr` doesn't have the index command 
```
$ expr index --with-netcdf=sfd =
expr: syntax error
```

rmcgibbo:
Resolves #216. ::grumble 1000 line bash script, grumble::

### DIFF
--- a/configure
+++ b/configure
@@ -397,7 +397,7 @@ VALUE=""
 while [[ ! -z $1 ]] ; do
   # If arg is Key=Value, separate into Key and Value
   VALUE=""
-  POS=`expr index "$1" '='`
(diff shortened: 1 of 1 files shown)


swails:
:+1:

PR #226, opened 2016-01-12
Do not require ncdump to come from AmberTools in the bundled build

swails:
If any users specified `--with-netcdf /path/to/netcdf` in their Amber configure,
then ncdump will need to be taken from the PATH, not AMBERHOME/bin.  As I
(almost) exclusively use --with-netcdf, I get tons of errors like "ncdump not
found".


drroe:
> ncdump will need to be taken from the PATH, not AMBERHOME/bin

D'oh - should've considered that...thanks!

swails:
Hindsight's 20-20 :).  It was nice that this was a 3-line fix for every test, though...

PR #228, opened 2016-01-22
Add parallel trajectory processing support

drroe:
This major PR gives cpptraj the ability to process trajectories in parallel at the individual level. This means that multiple threads can be assigned to read a single trajectory, or in ensemble processing multiple threads per ensemble member can be used. This required a major overhaul of how parallelization is handled; everything is now handled by the classes Parallel (static), Parallel::Comm, and Parallel::File instead of the collection of individual routines that was in MpiRoutines.c.

Everything is supported with the following caveats:
1. The clusterdihedral, contacts, spam, density, gist, stfcdiffusion, lipidorder, createreservoir, and pairwise actions cannot be trajectory parallelized (they still work with ensemble parallel, 1 thread per
member).
2. Any unformatted output during trajectory processing (e.g. from 'jcoupling outfile' etc) does not work (threads will overwrite each other). All formatted (i.e. data set/data file) output works.
3. COR, binpos, mol2, pdb, conflib, sqm input, sdf, and tinker trajectory output cannot be trajectory parallelized (still works with ensemble parallel, 1 thread per member). All trajectory reads should work fine in parallel.
4. Any trajectory output where cpptraj doesn't know everything about what the final trajectory will look like (outtraj maxmin, trajectory output after filter, trajout framerange, trajout append, check ‘skipbadframes’) is not supported.
5. Changing topology during run not supported.
6. In runs after the first, any data sets created by e.g. analyses are not broadcast to all threads, so they won't be picked up. Some things will work, some won't (the first run should always work).
7. Any actions following things like ‘check skip’ where frames can be skipped, data sets can end up being the wrong size.
8. Analyses, crdaction, and crdout are run using a single MPI thread (can still take advantage of openmp though).
9. No interactive mode.

In addition, OpenMP + MPI is now supported.

This PR also upgrades the testing framework so it can be used in parallel, adds more tests and enhances some existing tests.


swails:
Wow... nice work.

hainm:
crazy update. :+1:

drroe:
Thanks! It's been a long time coming...

PR #232, opened 2016-01-23
Changes to enable compile with parallel netcdf in AmberTools

drroe:
Use variables that will be set by AT configure.


drroe:
Since this commit only affects the AT Makefile, merging before appveyor test finishes.

PR #236, opened 2016-01-29
Improvements to configure, timing, and parallel ensemble processing

drroe:
This PR covers various improvements, enhancements, and clean-ups.

First, configure now tests parallel NetCDF if one is specified. It also only uses the INCLUDE variable when testing instead of all flags (which should not be necessary).

Second, various kinds of output have been improved. The biggest change to output is that run timings have been made more uniform between the different run types and are reported at the end of a run (some timings like FPS are still reported within the run). Various debug and output messages have been cleaned up and improved, particularly during ensemble runs.

Third and probably most important, the efficiency of ensemble setup in parallel has been greatly improved. Previously all threads would all run the same setup on all files in the ensemble, even though they were only responsible for reading one. This became particularly inefficient when multiple threads per ensemble member were used. Now each member of the ensemble in parallel only sets up the member it is responsible for. This results in great speedup, particularly for large ensembles and high processor counts (on BW got 2 orders of magnitude speedup for ensemble setup this way).

### DIFF
--- a/configure
+++ b/configure
@@ -89,7 +89,7 @@ CompileError() {
 # Test compile test program testp.cpp
 TestCxxProgram() {
   echo "$1"
-  COMPILELINE="$CXX $CXXFLAGS -o testp testp.cpp $2"
+  COMPILELINE="$CXX -o testp testp.cpp $2"
   $COMPILELINE > /dev/null 2> compile.err
   ./testp | grep "Testing" > /dev/null
   status=$?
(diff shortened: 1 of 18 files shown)


drroe:
@hainm I tested this with pytraj and it seems to build OK. Let me know if you try this PR and see otherwise.

hainm:
I am running tests now. So far seems ok.

hainm:
passed.

Issue #223, opened 2016-01-08
Triclinic distance bug


rmcgibbo:
I think I found a bug in the distance computation for some very skewed unit cells.

To reproduce, consider the following structure:
 ```
$ cat structure.pdb
CRYST1   20.000   20.000   20.000  41.55  56.39  17.52 P 1           1
MODEL        0
ATOM      1  O   HOH A   1       8.029   5.236   5.067  1.00  0.00          O
ATOM      2  H1  HOH A   1      13.771   5.666  14.130  1.00  0.00          H
TER       3      HOH A   1
ENDMDL
END
```

`cpptraj` gives a distance of 7.59A between these two atoms.

```
$ echo 'distance @1 @2 out distance.txt' | cpptraj  -y structure.pdb -p structure.pdb > /dev/null; cat distance.txt
#Frame      Dis_00001
       1       7.5932
```

However, I think I can find a closer distance between the two atoms. Here's my code:
```
import numpy as np
import scipy.spatial

# unit cell vectors
a = np.array([20., 0., 0.])
b = np.array([19.0720403, 6.0214001, 0.])
c = np.array([11.0718369, 14.6431303, 7.9368222])

# coordinates
x0 = np.array([8.029, 5.236, 5.067])
x1 = np.array([13.771, 5.666, 14.13])

def angle(v1, v2):
    return np.rad2deg(np.arccos(np.dot(v1, v2) / (np.linalg.norm(v1) *
                                                  np.linalg.norm(v2))))

# confirm that these unit vectors match the unit cell lengths and angles
# in the PDB file
print('alpha angle ', angle(b, c))
print('beta angle  ', angle(a, c))
print('gamma angle ', angle(a, b))
print('lengths     ', [np.linalg.norm(v) for v in (a, b, c)])
print()

offset_1 = (-2 * a) + (2 * b) + (-1 * c)
print('distance (-2,2,-1)  ',
      scipy.spatial.distance.euclidean(x0, x1 + offset_1))

offset_2 = (-4 * a) + (5 * b) + (-2 * c)
print('distance (-4,5,-2)  ',
      scipy.spatial.distance.euclidean(x0, x1 + offset_2))
```

Running this, I see that using `(-2,2,-1)` translations of the unit cells gives the distance reported by cpptraj, but using `(-4,5,-2)` translations gives a shorter distance.
```
$ python wut.py
alpha angle  41.5534977768
beta angle   56.38622
gamma angle  17.5218818781
[...]

PR #239, opened 2016-02-01
Check for skewed non-orthogonal box

drroe:
If box is too skewed to perform accurate imaging with current imaging routines, it is disabled.

@rmcgibbo If you get a chance would you mind checking this out? I tested on the system you originally provided and it seemed to work OK.

### DIFF
--- a/src/Box.cpp
+++ b/src/Box.cpp
@@ -202,6 +202,22 @@ void Box::SetBoxType() {
       mprintf("Warning: Low precision truncated octahedron angles detected (%g vs %g).\n"
               "Warning:   If desired, the 'box' command can be used during processing\n"
               "Warning:   to set higher-precision angles.\n", box_[4], TRUNCOCTBETA_);
+  } else if (btype_ == NONORTHO) {
+    // Check for skewed box.
+    const double boxFactor = 0.5005;
+    double Xaxis_X = box_[0];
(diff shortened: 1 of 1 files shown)


drroe:
Should address #223

swails:
I think better behavior is to keep the box on and just warn that computed distances *might* not be the absolute minimum.  The distances will be correct for any atom that lies within the radius of the largest sphere that can be circumscribed within the unit cell.  You could even compute what that distance would be and say something like "any distance below XYZ will be correct, other distances may not be the closest image because the unit cell is too skewed"

drroe:
> I think better behavior is to keep the box on and just warn that computed distances might not be the absolute minimum. 

I could potentially get on board with that. My only hesitation is that when most Actions perform imaging, cpptraj is *supposed* to adhere to the minimum image convention. When the box is skewed like this there is no guarantee that it can do this, so even though you're returning a closer distance it's not always going to be the *minimum* distance, and so technically it's wrong. In other words I'm not sure that returning an arbitrarily closer distance is better than just saying "well I know I'm going to fail sometimes so I won't even try", which seems more honest.

Although truth be told this is such a corner case that my feeling is that it won't matter for 99.9% of users...

drroe:
@swails OK I've come around to your way of thinking - give the user info and let them decide what to do. Otherwise the box info is always automatically removed which is bad.

Issue #243, opened 2016-02-02
dihcovar test


hainm:
likely that cpptraj does not have a test for dihcovar.

```bash
cpptraj/test (master) $ grep dihcovar Test_*/Run*
```
(return nothing)

drroe:
It does, but not in this repo. I'll work on adding it.

drroe:
Addressed by #245

Issue #145, opened 2015-10-29
travis test for cpptraj MPI


hainm:
we don't have yet. Just put a holder here when Dan finishes his MPI stuff with traj.

hainm:
bump this issue since @drroe  finished his stuff with parallel https://github.com/Amber-MD/cpptraj/pull/228

Need to add building for cpptraj.MPI on travis too.

drroe:
One small issue with this is that I don't think there is a parallel netcdf package available. The tests will still run but coverage will be a bit reduced.

hainm:
Is parallel netcdf required for all Actions or for only writing trajectory?

drroe:
Just a few tests. I just need to protect them against not having parallel netcdf and things will be ok.

drroe:
Addressed by #245

Issue #109, opened 2015-10-09
Have `-y` CL flag accept arbitrary number of arguments


swails:
This would allow shell globbing to work as expected.  I'll have a go at this when I get a chance early next week.

swails:
cc discussion in #39

drroe:
Addressed by #245

Issue #100, opened 2015-10-03
reference tag and name


hainm:
ack,

I've just realized that `[REF]` is not a name. I need to explicitly use ``[ ]`` to tag. Now I see why @drroe mentioned `tag` and `name` before.

Currently, I can not specify reference frame by name, right?

For example, this one should work (cpptraj man)
```bash
trajin Test1.crd
reference 1LE1.NoWater.Xray.rst7 [xray]
reference Test1.crd lastframe [last]
reference Test2.crd 225 [open]
rms Xray ref [xray] :2-12@CA out rmsd.dat
rms Last ref [last] :2-12@CA out rmsd.dat
rms Open ref [open] :2-12@CA out rmsd.dat
```

but this one does not, right? (I've tried and got error reference not found)
```
trajin Test1.crd
reference 1LE1.NoWater.Xray.rst7 xray
reference Test1.crd lastframe last
reference Test2.crd 225 open
rms Xray ref xray :2-12@CA out rmsd.dat
rms Last ref last :2-12@CA out rmsd.dat
rms Open ref open :2-12@CA out rmsd.dat
```

this is the reason why I was stuck in this issue
https://github.com/Amber-MD/cpptraj/issues/35

hainm:
ideally, I wish we can add `name` 

trajin Test1.crd
reference 1LE1.NoWater.Xray.rst7 ``name xray``
reference Test1.crd lastframe ``name last``
reference Test2.crd 225 ``name open``
rms Xray ref ``xray`` :2-12@CA out rmsd.dat
rms Last ref ``last`` :2-12@CA out rmsd.dat
rms Open ref ``open`` :2-12@CA out rmsd.dat

hainm:
oh, I just try to make a name for reference in Action_NativeContacts and refer its name when doing the action. It works. 

But I still want to hear your over idea when you have time.

drroe:
The "tag" functionality is from a time when topologies and reference structures were stored in their own separate arrays. It's still supported for backwards compatibility. I'll consider adding the `name` keyword to `reference` and `parm` - it's not a bad idea.

hainm:
> I'll consider adding the name keyword to reference and parm - it's not a bad idea.

I think this will be very convenient for pytraj. pytraj can take the reference from DSL. See my desired command below.

```python
        traj = pt.iterload("./data/md1_prod.Tc5b.x", "./data/Tc5b.top")

        commands = ['rmsd @CA',
[...]

PR #245, opened 2016-02-04
Numerous fixes and enhancements

drroe:
This PR covers a lot of ground.

- Address #96 for IRED output labels.
- Address #100; add `name` keyword for `parm` and `reference` commands which can be used in place of tags.
- Correctly handle file name wildcards on the command line for `-p`, `-y` etc ( #109 ). 
- Add `-d` and `-w` command line flags for reading and writing data sets from/to files (which also enables command-line conversion of data files).
- Add dihedral co-variance matrix tests ( #243 ), as well as tests for `curvefit`, `multicurve`, `avg`, and `vectormath`.
- Add data set scalar math functionality for 2D and 3D data sets (min, max, avg, etc).
- Add `allframes` keyword to `nastruct` to allow finding of base pairs every frame (not parallel-enabled yet).
- Fix issue where in `nastruct` step parameters were being calculated even if base pair parameters were not - now correctly set to 0.0 when this is the case.
- Add `calcnohb` keyword to calculate parameters even if base pairing not detected (no hbonds).
- nastruct: Add BP to base pair output, which is 1 if base pair is present (number hbonds > 0) and 0 otherwise. Also add Zp to base pair step output.
- Fix `nastruct` in parallel when base pairing changes in a run.
- Attempt to get tests working with MPI in Travis ( #145 ).
- `avg` analysis now uses data set/data file framework.
- Allow either vector to be of size 1 in `vectormath`, which allows vec OP vec(n) math.
- Add `nomod` keyword to `rms` command to prevent modification of coordinates during calculation of RMSD and adds a test for it.


drroe:
OK, I clearly don't know enough about Travis. I'm having all kinds of weird issues. The current problem is 
```
E: Unable to locate package openmpi
```
But it seems like openmpi is on the whitelist:
https://github.com/travis-ci/apt-package-whitelist/blob/7a5250869012848f35bc16f5d0572749bbc84e39/ubuntu-precise#L9121-L9129

I'm going to keep at this for a little bit longer but if I can't figure it out I'll just disable the Travis MPI build for this PR.

hainm:
@drroe  can you check this?

https://github.com/mpi4py/mpi4py/blob/master/conf/travis/install-mpi.sh

mpi4py use MPI on travis.

drroe:
Hmm - removing the `openmpi` package did seem to do the trick - build and most tests succeeded. The ones that failed all do so with:
```
[testing-worker-linux-docker-2b86ccd2-3381-linux-1:10951] *** An error occurred in MPI_Reduce
[testing-worker-linux-docker-2b86ccd2-3381-linux-1:10951] *** on communicator MPI COMMUNICATOR 3 SPLIT FROM 0
[testing-worker-linux-docker-2b86ccd2-3381-linux-1:10951] *** MPI_ERR_ARG: invalid argument of some other kind
[testing-worker-linux-docker-2b86ccd2-3381-linux-1:10951] *** MPI_ERRORS_ARE_FATAL (your MPI job will now abort)
```
The failing tests are Test_AtomicCorr, Test_Grid (grid generation from bounds), and Test_Matrix. I haven't seen this error before with mvapich2 or mpich, so I'll have to try to reproduce tomorrow.

drroe:
OK, I think I have fixed the OpenMPI issues. Turns out OpenMPI gets cranky if MPI_Reduce() is called with a null pointer, even if the count is 0 (this isn't an issue with mpich/mvapich). Should be fixed now (I hope).

drroe:
Now there is a new failure in Test_Cluster_TrajWrites. It didn't appear in the previous build and I can't reproduce it locally (trying to replicate the Travis environment as close as possible; GNU 4.6.2 and OpenMPI 1.4.5). I'll re-trigger the build and see if the same error happens. If it does, this may have to be one of the cases that's allowed to fail for now.

drroe:
OK, at long last the Travis build seems OK. I'll have to keep an eye out for transient failures in `Test_Cluster_TrajWrites`, which may indicate some weird subtle bug somewhere, but for now this seems good to merge (finally).

hainm:
@drroe  are you considering change `nomod` (rms) to something easier to remember?

I chose `update_coordinate` in pytraj.

swails:
Mr. "let's change `coordinates` to `xyz` to save typing"?? `nomod` -> `update_coordinates`?

What have you done with Hai? :smiley:

hainm:
:P 

This is my rational.

`xyz` commonly used in some other python packages (including mdtraj, ...). And people seems oftenly use `xyz` (traj.xyz), so I would like to keep it to make it short. 

For example, in this nglview, it have similar code wrapping for mdtraj and pytraj (IMO, it's better to share similiar API): https://github.com/arose/nglview/blob/master/nglview/__init__.py#L148-L193

Per `update_coordinates` in `pytraj.rmsd`, I don't think this will be frequently used (including myself), so I leave it that way. (`update_xyz=True` seems vague to me).

hainm:
but  people can change their mind too =))

I used to suggest to use `as_structure=True` for parmed but after using `parmed` for a while, I prefer to use `structure=True`, look prettier.

drroe:
> are you considering change nomod (rms) to something easier to remember?

`nomod` = "no modify", which I don't think is that hard to remember. It's also consistent with the other option keywords `nofit` and `norotate`, so I will leave this as-is.

hainm:
nomod -> nomodify? (norot -> norotate) :D

hainm:
anyway, just minor thing.

Issue #27, opened 2015-07-30
rotate molecule with axis = atom1 - atom2 


hainm:
@drroe 

I am not sure how to use "rotate" in cpptraj to do this "Rotate the molecule by an amount (in degrees, specified by -d) along the axis defined by two atoms (specified by -a1 and -a2)" [from antechamber](http://ambermd.org/antechamber/ac.html)

Does cpptraj support this? thanks

cpptraj: https://github.com/Amber-MD/cpptraj/blob/master/src/Action_Rotate.cpp#L9-L10

drroe:
There isn't an action that supports this, but you can do it with the Frame, Vec3, and Matrix_3x3 classes (following example assumes MyFrame already has coordinates):

```C++
// a1 and a2 are integers representing atom indices (from 0)
Vec3 axisOfRotation = MyFrame.SetAxisOfRotation(a1, a2);
// theta is a double, must be in radians
Matrix_3x3 U;
U.CalcRotationMatrix(axisOfRotation, theta);
MyFrame.Rotate( U ); // Rotates all coords
```
Do you need this to be an Action?

hainm:
@drroe I don't need extra action for my need since I can follow above code. But it's great if you can introduce to cpptraj. I think cpptraj has already some actions that can change coordinates, we can add more (for example, I really like Action_Makestructure).

drroe:
@hainm This is something that could be introduced as e.g. a keyword to the Rotate action, but I'm curious as to why you would ever want to do this. Do you have specific applications in mind?

hainm:
@drroe I was thinking about rotating a group of atoms along a defined axis, translating it to a pocket. In the future, we can extend pytraj/parmed/cpptraj for modelling. 

I am playing with force field, so I would like to use `cpptraj` to rotate some dihedral angles then calling `pysander` to evaluate energy. I am too lazy to look at tleap.

drroe:
Does the 'makestructure' action not do what you want? Or have you been
having trouble getting it to work? I'm not able to access my files easily
right now but I can get you some examples tomorrow.

-Dan

PS - Related, I will have to start adding more test cases (like you've
requested) since they do serve as examples.

On Tuesday, August 4, 2015, Hai Nguyen <notifications@github.com
<javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:

> @drroe <https://github.com/drroe> I was thinking about rotating a group
[...]

PR #247, opened 2016-02-05
Add option to 'rotate' command to rotate around user-defined axis

drroe:
Addresses #27. Also adds a test.


hainm:
@drroe 

Do the names 'axis0', 'axis1' sound right? (given that axis is axis, not a point).

PR #249, opened 2016-02-08
Add option to link to Sander API

drroe:
Cpptraj can now be configured with `-libsander` to link to the Sander API if available. This also adds a new Action, `esander`, which can be used to calculate energy via the Sander API and store and write the data via Cpptraj's DataSet/DataFile framework. Among other things this means users will be able to parallelize an `imin = 5` style calculation by splitting up the trajectory among threads. Adds two tests: one for PME and one for GB. However, these tests will not currently be run via CI since libsander is required.

This also adds the ability to read EPB, ECAVITY, and EDISPER energy terms from MDOUT files.


hainm:
> Cpptraj can now be configured with -libsander to link to the Sander API if available. 

One thing I really like about Python as glue language is that I can install package independently. For example, if I want to use sander with pytraj, I just need to install pysander without reinstall pytraj again. :D

swails:
You can do that in C++ pretty easily, too.

hainm:
> You can do that in C++ pretty easily, too.

:D can you give me a link to read (not sure how to google)

and does cpptraj should not do that? For example: If I let user to install pytraj, cpptraj will be installed too (but not link to sander API). But later, user decides to use sander (in pytraj) via cpptraj, can cpptraj offer that without recompiling libcpptraj.so?

drroe:
> can cpptraj offer that linking without bother user to recompile libcpptraj.so?

Currently no, but that's just because of the way the library is currently built (as a giant monstrosity of all functionality). A better way to implement it would be to have cpptraj components in different libraries, which you could then link as wanted (so the sander API stuff could be in libcpptrajsander.so or something and then you would link with -lcpptraj -lcpptrajsander). However, changing the way things are currently done to achieve this would be a lot of work for hardly any benefit in my opinion. 

What I eventually envision is that once the sander API linkage is tested and more stable it will be on in AmberTools by default, and *maybe* on in GitHub if $AMBERHOME is set and libsander.so is found and valid.

swails:
> :D can you give me a link to read (not sure how to google)

Google `dlopen`.  It's actually used inside sander for PLUMED for exactly this purpose.  i.e., PLUMED will work with sander as soon as you compile PLUMED and install it somewhere that ld can find it (i.e., either in some place referred to by `/etc/ld.so.conf` or somewhere in `LD_LIBRARY_PATH`).

OpenMM uses the same thing so that you can simply compile a plugin against OpenMM and it will automagically work without having to recompile all of OpenMM over again.  Or install a compatible CUDA and the CUDA Platform will just start working, without having to recompile all of OpenMM.

It's not nearly as easy as Python's import machinery, but it's also not particularly difficult (and is in use even in Amber :wink:).

hainm:
awesome, I will read more about that.

swails:
One thing we could also do is unify some of this testing framework a little.  I plan on making a pysander conda package, which means that if you load miniconda onto Travis, you can install libsander as simply as typing 'conda install pysander', then link to the libraries that conda installs.

I'll try and get a conda package built this week before my new job starts.  It would also provide a platform to run the cpptraj and pytraj tests at the same time, if that's something you guys wanted to pursue at some point (maybe it'll be an allowed failure on the cpptraj build just to give a heads-up when the pytraj build is broken).

hainm:
> conda install pysander

thanks. This is exactly I have been dreaming about. (I turn off code coverage for pytraj.energy_decomposition on travis since there is no pysander).

PR #250, opened 2016-02-09
Always create temporary Amber topology for libsander

drroe:
Since the current API requires a valid Amber topology, have cpptraj always write one out since it's not always guaranteed the topology file exists. This can happen if you `strip` atoms for example. This PR also adds an `install_libsander` target to AmberTools - not yet part of the main build but is suitable for testing.


hainm:
@drroe does cpptraj still evaluate energies even if user strip to weird Topology?

swails:
I don't see why not... ParmEd supports such things.

drroe:
> does cpptraj still evaluate energies even if user strip to weird Topology?

It should, yes. As long as the topology is considered valid by sander it will work.

PR #252, opened 2016-02-10
Add a Mac build to the Travis matrix.

swails:


hainm:
II've been trying too by using Miniconda for netcdf but still got problem with `lgfortran`

https://travis-ci.org/hainm/cpptraj/builds/107345858

swails:
You can use `brew` to install the necessary prerequisites -- no need for conda (yet).  Although if we're going to start using conda for libsander we might as well...

swails:
OK, this is done!  It works!

The failure here is the set of real failures that I'm seeing on my Mac.  Have a look at the failures if you want to.

hainm:
I tried pytraj on osx with @swails mactravis and the installation went well for both libcpptraj.so and pytraj.

However, when run pytraj, I got 

```
ImportError: dlopen(/Users/travis/miniconda/envs/myenv/lib/python3.5/site- 
packages/pytraj/cpp_options.cpython-35m-darwin.so, 2): Library not loaded: libcpptraj.so 
  Referenced from: /Users/travis/miniconda/envs/myenv/lib/python3.5/site- 
packages/pytraj/cpp_options.cpython-35m-darwin.so 
  Reason: image not found
```

Do we need to add another flag (rather -fPIC) on osx?
(http://stackoverflow.com/questions/14173260/creating-shared-libraries-in-c-for-osx)

swails:
I think -fpic should still be necessary.  Regardless, it's not *hurting* anything (it may be that -dynamiclib implies PIC, or that Macs build that by default).

The error you're seeing has to do with a bungled RPATH, and that's currently worked around in Amber by setting DYLD_FALLBACK_LIBRARY_PATH (which has a similar task as LD_LIBRARY_PATH in Linux).  That's a less reliable route going forward, so you may want to look into something like `install_name_tool` to set the RPATH correctly.

hainm:
Thanks. I will try.

drroe:
OK, it looks like the `nativecontacts` sorting issue has been fixed. The `gist` results definitely seem like round-off to me, but I'm not sure about the curve fitting. Since I still don't have access to OSX my options for debugging are limited. We could just allow these tests to fail on OSX. What's your opinion on how to proceed?

swails:
It all looks like round-off.  Do you use dacdif for the comparisons?  How hard would it be to be a little less strict on these tests?  Another difference between Macs and the Linux builds is that the math libraries come from the mac Accelerate framework (as opposed to blas and lapack).  Are any of those routines used in curve fitting and gist?  Because if so, that could explain the differences (and would be worth "fixing" so the tests are more likely to pass with other substitutions for the math libraries, like MKL or GOTO).

drroe:
> Do you use dacdif for the comparisons? How hard would it be to be a little less strict on these tests?

No, all comparisons are straight up. Much faster that way and I like to be strict :-). The strategy so far is to let tests that are known to fail on certain OSs/build environments (e.g. windows, MPI) get off with a warning but still print the diffs so that they can be examined. I suppose a case can be made for implementing an infrastructure that can say "yep, that's round-off" but personally I like leaving that up to the user.

>  Another difference between Macs and the Linux builds is that the math libraries come from the mac Accelerate framework (as opposed to blas and lapack). Are any of those routines used in curve fitting and gist?

No, but they are both complicated beasts so the round-off doesn't surprise me. I've seen some differences in the curve fitting test before depending on the version of compilers (mostly intel), but it hasn't been so huge that I've bothered to track it down. It's possible there's an uninitialized or improperly initialized variable somewhere that could be responsible. I'll try to look into reproducing the curve fit diffs with intel compilers today and see if I can find something. GIST I really don't want to dive into - it's not my code and the differences are so small that I'm OK letting it fail on OSX (gist.out is currently allowed to fail on windows). The allowed failures mechanism will have to be updated to allow more than one OS to fail on any given test but that's something I can do pretty easily.

hainm:
@swails 

I tried to use DYLD_FALLBACK_LIBRARY_PATH but got segmentation fault

`$ if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export DYLD_FALLBACK_LIBRARY_PATH=`pwd`"/cpptraj/lib":$DYLD_FALLBACK_LIBRARY_PATH else export LD_LIBRARY_PATH=`pwd`"/cpptraj/lib":$LD_LIBRARY_PATH; fi`

https://travis-ci.org/Amber-MD/pytraj/builds/108714859#L6207-L6225
```
$ python ./run_simple_test.py
/Users/travis/build.sh: line 45:  3574 Segmentation fault: 11  python ./run_simple_test.py
```
really need to buy a Mac for testing. :D

drroe:
Incorporated via #272. Thanks for getting the ball rolling @swails!

PR #257, opened 2016-02-11
Improvements to sander API use by CPPTRAJ

drroe:
This PR changes cpptraj's standalone configure so it automatically tries to use the sander API if it is specified with `--with-sanderlib` or is present in `$AMBERHOME`. This can be disabled via the `-nolibsander` flag. If the sander API cannot be used a warning is issued but configuration is allowed to complete.

This also changes the AmberTools Makefile so it makes use of new variables related to the sander API that will be introduced to Amber's config.h:

- SANDERAPI_LIB: The sander API library
- SANDERAPI_DEF: The sander API define
- SANDERAPI_DEP: The sander API dependency

These will automatically be set by AmberTools configure for all builds except parallel. In parallel, configure will check that `libsander.so` is present and can be linked to; if not, linking to the sander API is disabled and a warning is issued but configuration is allowed to complete.

### DIFF
--- a/Makefile_at
+++ b/Makefile_at
@@ -1,15 +1,13 @@
-# CPPTRAJ - Main Makefile
+# CPPTRAJ - AmberTools Makefile
 # Daniel R. Roe
 # 2010-11-18
-# Revised 2015-03-09
+# Revised 2016-02-11
+include ../config.h
 
 # Create standalone cpptraj binary in ./src/
-all:
-	cd src && $(MAKE)
+all: install
 
-# Create standalone cpptraj binary
-install_local:
-	cd src && $(MAKE) install
+install: $(INSTALLTYPE)
 
 # Create cpptraj/ambpdb binaries within AmberTools
 serial:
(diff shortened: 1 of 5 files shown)


hainm:
Can we not build with libsander for libccptraj since pytraj uses pysander.

hainm:
Or may be pytraj just use -nolibsander flag.

But my idea is to make minimal dependeces as possible. Sometimes thing goes crazy. (For example, pytraj looks for wrong libcpptraj dir).

hainm:
There is only one issue left 

https://github.com/Amber-MD/pytraj/issues/981

drroe:
> Can we not build with libsander for libccptraj since pytraj uses pysander.

I can make it so that with the standalone configure, `-shared` implies `-nosanderlib`.

In AmberTools we can re-make libcpptraj.so without the SANDERAPI flags when it is needed. Since only 3 objects currently need to be remade this is actually not that expensive.

PR #258, opened 2016-02-12
Ensure libsander.so not included in libcpptraj.so

drroe:
This is done via `configure` for standalone cpptraj and via a modified Makefile rule under AmberTools. The source directory is also now cleaned after configure if needed (including libreadline.a) to avoid any compile issues with e.g. PIC compiled code with non-PIC code etc.

### DIFF
--- a/Makefile_at
+++ b/Makefile_at
@@ -21,9 +21,12 @@ openmp:
 parallel:
 	cd src && $(MAKE) -f Makefile_at install_mpi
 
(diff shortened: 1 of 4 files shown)


hainm:
tested with stand alone and AT versions. Works for me.

drroe:
> tested with stand alone and AT versions. Works for me.

Hallelujah

Issue #256, opened 2016-02-11
add option to keep extension 


hainm:
as we discused on gmail, I prefer to have an option to keep the extension.

`.rst7.1` --> `.1.rst7`
`.pdb.1` --> `.1.pdb`

Rationale: Many programs recognize files based on extension. So don't make it hard for user (to write his/her own renaming script).

PS: and `.1.rst7` looks prettier (IMO).

drroe:
Hopefully addressed by #259

drroe:
Done.

hainm:
@drroe what do you think about adding 'keepext' for clustering too?

hainm:
updated ` keepext for 'mask' command?`

PR #259, opened 2016-02-12
Minor fixes

drroe:
- Fixes a weird segfault that I saw randomly sometimes involving std::locale.
- Addresses #256


hainm:
tested with pytraj for rst7 ext and works fine.

PR #266, opened 2016-02-22
Fix shared library handling/creation on OS X

drroe:
Previously cpptraj configure always used the '.so' suffix for shared libraries, so detection could fail on platforms where this is not the case. Detection is now fixed and libcpptraj will be generated with the appropriate suffix for the platform.


hainm:
related PR: https://github.com/Amber-MD/cpptraj/pull/252

PR #272, opened 2016-02-25
Upgrade test framework to handle round-off

drroe:
This PR adds Nelson H.F. Beebe's ndiff.awk script (taken from AmberTools 16) to the test framework, allowing certain tests a degree of numerical tolerance to account for round-off errors. It gets rid of the `allowfail` and `parallelfail` DoTest keywords, which would report failures as warnings. The new framework is much more robust and is now consistent with what Amber does.

With the new framework, it should now be possible to test on Travis-CI using OSX, so @swails changes from PR #252 are incorporated into this (thanks to Jason Swails for getting that running in the first place).

This PR also fixes the summation of the histogram in the `pairdist` command in parallel.


hainm:
> Upgrade test framework to handle round-off
...
> This PR also fixes the summation of the histogram in the pairdist command in parallel.

may be in the future, you should make different PRs for different issues?

This is also easier for me to follow what changes in cpptraj.

drroe:
> may be in the future, you should make different PRs for different issues?

I considered it part of the same issue since it resulted in a test failure due to round-off (but only in parallel).

hainm:
> I considered it part of the same issue since it resulted in a test failure due to round-off

ah, agree.

drroe:
Woo! Finally got OSX working!

PR #276, opened 2016-02-27
Try to add libsander support to Travis testing.

swails:


swails:
So it's close.  It's missing some libfftw3 symbols and libnetcdf symbols, but the files are getting onto Travis and they're OS-compatible, so that's good.

drroe:
@swails What do you think of having the option to build libsander.so so it has no external dependencies? Seems like it could potentially solve this issue and would be nice to have on cygwin where building a DLL with external dependencies can be a bit tricky.

swails:
That's the current intended behavior. The only reason the dylib isn't standalone right now is because I used an external NetCDF which links the dylib instead of the .a. It's on my to-do list and part of what I hope to get working in this PR

swails:
I still have to figure out the missing fftw symbols.

swails:
OK.  I rebuilt the conda package for Linux/pysander after building Amber with the bundled NetCDF (required to make a standalone libsander.so file).  I also adjusted the FFTW3 linking to link the static library instead of the dynamic one.  This should eliminate the missing FFTW3 symbols as well.

Not sure what's going on with the Mac VM, so I'm trying to debug that one.  But hopefully the Linux builds will work *and* test esander.

swails:
The Linux builds are working! https://travis-ci.org/Amber-MD/cpptraj/jobs/112437052#L392

Mac is still failing (expectedly though)

swails:
OK, looks like mac is failing for the same reason.  Rebuilding the conda package now which should hopefully fix it.  This should be good to go very soon!

swails:
OK, this should be good to merge once it passes.  I took out the sander API testing on macs, since it's a rabbithole and I don't have time to find my way around it right now.

At least the API is being tested now, so this is a lot better than nothing (and it also tests *not* using the sander API, which is probably a good thing as well).

swails:
OK, Travis is passing now.  This is ready for merging.  You can look at the Travis logs, too -- the sander tests *are* being run.

drroe:
@swails Outstanding work - thanks Jason

hainm:
Can't wait to try. Thanks Jason.

PR #288, opened 2016-03-11
Frame(frame) should always return a new Frame with its own memory

hainm:
this fixes https://github.com/Amber-MD/pytraj/issues/1166

### DIFF
--- a/src/Frame.cpp
+++ b/src/Frame.cpp
@@ -148,7 +148,7 @@ Frame::Frame(const Frame& rhs) :
   V_(0),
   F_(0),
   remd_indices_(rhs.remd_indices_),
-  Mass_(rhs.Mass_),
+  Mass_(rhs.Mass_)
 {
   // Copy coords/velo/forces; allocate for maxnatom but copy natom
   int maxncoord = maxnatom_ * 3;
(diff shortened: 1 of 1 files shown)


src/Frame.cpp
```diff
@@ -149,7 +149,7 @@ Frame::Frame(const Frame& rhs) :
   F_(0),
   remd_indices_(rhs.remd_indices_),
   Mass_(rhs.Mass_),
-  memIsExternal_(rhs.memIsExternal_)
+  memIsExternal_(false)
```
drroe: This feels correct - a copy constructor should indeed make a copy. However you may want to benchmark and see if there are any places where this hurts performance. There may be cases where we want to just pass the address of `X_` etc - in such cases we can create a new function to do that. Also, can you get rid of the `if (memIsExternal_)` part? If `memIsExternal_` is always false then it will never get called.

hainm:
> However you may want to benchmark and see if there are any places where this hurts performance. 
Do you mean performance in pytraj? pytraj does not use copy constructor much.

> can you get rid of the if (memIsExternal_) part? If memIsExternal_ is always false then it will never get called.

updated.

hainm:
phew, green now.

drroe:
OK looks good. Thanks

hainm:
image

PR #291, opened 2016-03-18
Add 'bigendian' option to TRR trajectory output

drroe:
Gromacs analysis tools appear to require TRR files be written big-endian, so add `bigendian` option to TRR trajectory output to force big-endian. Also adds a `dt` option for setting time step factor.

### DIFF
--- a/src/ByteRoutines.cpp
+++ b/src/ByteRoutines.cpp
@@ -49,3 +49,11 @@ void endian_swap8(void *x, long nswap) {
   }
 }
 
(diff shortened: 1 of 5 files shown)


hainm:
lol. quick.

hainm:
no test yet? :P

drroe:
I have a local test. I tend not to put trajectory tests here since the files are generally large and the test cases are already pretty big. Perhaps a small one down the road.

PR #294, opened 2016-03-27
Attempt to fix pytraj on OS X.

swails:
Building directly into $(LIBDIR) can be considered a temporary hack until we get
the install_name_tool working correctly. However, this has no effect on the
standalone configure script, and is useful for AmberTools since pytraj is
currently the only user of the cpptraj API.  This is needed to work toward
getting pytraj to work on Macs.

With just this change, pytraj works using the GNU compilers from MacPorts with a conda Python.

### DIFF
--- a/src/Makefile_at
+++ b/src/Makefile_at
@@ -26,8 +26,7 @@ install: cpptraj$(SFX) ambpdb$(SFX)
 	/bin/mv cpptraj$(SFX) $(BINDIR)/
 	/bin/mv ambpdb$(SFX) $(BINDIR)/
 
-libcpptraj: libcpptraj$(SHARED_SUFFIX)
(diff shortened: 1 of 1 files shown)


hainm:
This is cool, Jason. 

FYI: if I use conda to build libcpptraj and pytraj before, pytraj works. 

This is the receip with clang for libcpptraj
https://github.com/Amber-MD/conda-recipes/blob/master/libcpptraj/build.sh

It is likely that conda did some post processing to fix libc++ and libstdc++ as you mentioned in email.

This is the log on travis: https://travis-ci.org/hainm/TestPytraj/jobs/118634679

swails:
I think I'm very close to getting pytraj working for all compiler combos on OS X within Amber.  I'll submit a PR to the pytraj repo shortly.

drroe:
Good stuff @swails, thanks.

PR #295, opened 2016-04-06
Enable changing of matrix data set mode/type.

drroe:
This is necessary if e.g. you want to read in a previously generated matrix data set and run `diagmatrix`; if the type is not appropriately set the command will fail.


hainm:
welcome back. :P

PR #298, opened 2016-04-08
Improvements to 'comparetop'

drroe:
Much better and more stable sorting now results in reporting accurate differences (no more false positives). Also improve output for bonds/angles/dihedrals. Still needs more testing but almost good to go.

### DIFF
--- a/src/Exec_CompareTop.cpp
+++ b/src/Exec_CompareTop.cpp
@@ -49,6 +49,13 @@ static void PrintAngT(CpptrajFile& output, Topology const& parm,
                 *(parm[first.A3()].Type()));
 }
 
+/// Function for printing AngleParmType
+static void PrintAngP(CpptrajFile& output, Topology const& parm,
+                      AngleParmType const& first, char dir)
+{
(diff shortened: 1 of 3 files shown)


hainm:
Dan, what is useful case for this?

drroe:
> Dan, what is useful case for this?

You create a topology for a specific FF with AmberTools 15. Later on you create another topology for the same FF with AmberTools 16. You diff them and uh-oh; they don't match! What do you do? Go through each topology line by line and compare every single parameter yourself? I can feel the migraine beginning...

Now you don't have to! With `comparetop`, cpptraj will act like a `diff` command for topology files, pointing out exactly what's different between them. 

For the specific case of AT15 vs AT16 there are some zero-height torsion terms added by Jason to (I believe) ensure wildcards don't get used when they're not supposed to. An energy calc reveals they are identical; `comparetop` lets you know why and how.

(Aside: such functionality _probably_ belongs in parmed; but I don't know that quite so well and needed this to work quickly...)

hainm:
> I can feel the migraine beginning...

agree.

> such functionality probably belongs in parmed

yeah, parmed should definitely have it too. 

> needed this to work quickly

I see.

PR #321, opened 2016-05-16
Wavelet analysis feature extraction code

drroe:
This PR primarily adds code for wavelet analysis feature extraction (WAFEX), which is a way of filtering resultant wavelet maps to extract significant features from MD trajectories.

Also does the following:

- OpenMP-parallelize the wavelet convolution for each atom.
- Adds the ability to read data from Gnuplot files written by cpptraj.
- Adds the ability to read/write Gnuplot data in binary (much faster and compact than ASCII).
- Adds the ability to change the palette of output Gnuplot files.
- Adds more tests (wavelet, matrix).
- Adds a progress timer, which can be used when updates at specific time intervals are desired.


hainm:
It is time to add to pytraj. Thumb.

PR #328, opened 2016-05-20
Enhancements and fixes to Amber topology parsing and molecule determination

drroe:
This PR was inspired by trying to load in an ~11 million atom chromatin system and getting a segmentation fault. This is a complete rewrite of the Amber topology parser to be faster (about 1.8x over the old one) and more memory efficient (barely uses more memory than that needed for the Topology itself). In other words, this is how I should have written the parser when I changed around the Topology class a while back...

The segmentation fault was actually caused by blowing the stack from the recursive function which determines molecules. There is now a non-recursive version of this function that kicks in at ~150K atoms - below this the recursive search is still slightly faster. 

Bottom line is now cpptraj seems to be able to handle systems of at least 11599998 atoms (loads in about 32 seconds, though initial loads could be slower).


hainm:
download png

swails:
Do me a favor and keep this topology file away from parmed. 🙂

Issue #317, opened 2016-05-14
CIF parsing error: Internal Error: Attempting to add empty CIF data block.


hainm:
I was trying to read http://www.rcsb.org/pdb/explore.do?structureId=2b5t (`2b5t.cif`)

Got
`cpptraj -p 2b5t.cif`

```python
CPPTRAJ: Trajectory Analysis. V17.00b
    ___  ___  ___  ___
     | \/ | \/ | \/ | 
    _|_/\_|_/\_|_/\_|_

| Date/time: 05/13/16 21:10:32
| Available memory: 1.419 GB

	Reading '2b5t.cif' as CIF File
Error: Line 207: # of columns in loop data '_entity_poly' (1) < # column headers (7)
Error: Line 209: Data '', expected quote character.
Error: Line 211: Data '', expected quote character.
Internal Error: Attempting to add empty CIF data block.
Error reading topology file '2b5t.cif'
Error: Could not open topology '2b5t.cif'
TIME: Total execution time: 0.0013 seconds.
Error: Error(s) occurred during execution.
```

hainm:
`ParmEd` loads nicely
![2bt5](https://cloud.githubusercontent.com/assets/4451957/15265327/6d5eda18-194f-11e6-9d67-62b79c5ca469.png)

drroe:
Should be addressed by #331

hainm:
![screenshot](https://cloud.githubusercontent.com/assets/4451957/15512700/314e3718-21af-11e6-97d5-db4ccc5505b4.png)

yeah

PR #331, opened 2016-05-23
Fix parsing of CIF files.

drroe:
Data fields with quotes or semi-colons are now treated properly. Addresses #317.

Also improves Topology; when adding atoms, recognize change in residue insertion codes as beginning a new residue.


drroe:
Since the appveyor build never triggered on the initial PR, I made a few more commits to the branch. Now tests have been added for the CIF and PDB formats. Also, parsing of the B-factor and occupancy fields from PDB files has been improved. Finally, Test_Comprehensive was removed since it was large and redundant, and Test_LIE was changed so that it used existing topology/trajectory files in order to reduce the space it took.

Issue #320, opened 2016-05-16
RMSD: option not to dump data  to dataset?


hainm:
hi,

is there any option that only perform rms fit without dumping rmsd value to dataset?

My issue is: I want to use out of core TrajectoryIterator with NGLView and want to perform transformation in the fly: `rmsd`, `center`, `autoimage`... 

To make things easier, I use `ActionList` to store all actions, whenever iterating to get Frame, I call `ActionList` to do action on Frame (so all transformations will be done by cpptraj to avoid python overhead). However, cpptraj dumps rmsd data to DataSetList. This is ok for cpptraj context but it's not OK if I want to play the movie again and again.

Of course, I can reuse `Frame` translation and rotation methods but this can't not be combined this with `ActionList`.

Any suggestion?

hainm:
My current work around is: whenever the Dataset length reaches 1000 (frames), pytraj will remake `ActionList`, `DatasetList`, ...

drroe:
There is currently no option to do this. Adding one is possible, but if all you need is the RMS fit maybe it would be better to just have pytraj do it via `Frame::RMSD()` or one of the other RMSD functions - is that possible? Or does that involve the python overhead you mentioned in your initial post?

> perform transformation in the fly: rmsd, center, autoimage...

Just to make certain, you should image *before* performing an RMS fit. Also, `center` is redundant if you are using `autoimage` (since `autoimage` performs centering).

> My current work around is: whenever the Dataset length reaches 1000 (frames), pytraj will remake ActionList, DatasetList, ...

Couldn't you just remake the ActionList and DataSetList whenever you hit "play"?

hainm:
> Couldn't you just remake the ActionList and DataSetList whenever you hit "play"?

I want `pytraj` to be dependent from `nglview`. 

> Adding one is possible, but if all you need is the RMS fit maybe it would be better to just have pytraj do it via Frame::RMSD() or one of the other RMSD functions - is that possible? 

I want to use `RMSD` in `ActionList` so I can keep adding more actions.

```python
(traj.autoimage()
      .superpose()
      .translate('x 3.2')
      .scale('...'))
```

```python
actlist.add('autoimage')
actlist.add('rms')
actlist.add('translate')
...
for frame in traj: actlist.compute(frame)
``` --> much simple.

> Just to make certain, you should image before performing an RMS fit. Also, center is redundant if you are using autoimage (since autoimage performs centering).

[...]

Issue #319, opened 2016-05-15
missing test: velocityautocorr


hainm:

drroe:
Addressed  by #335

PR #335, opened 2016-05-25
Add action to perform only structure alignment

drroe:
Essentially a stripped-down version of the RMSD action that only performs fitting. Intended to address #320.


drroe:
Since the initial PR failed due to the CI tests not being properly triggered (again), add `velocityautocorr` test (addresses #319) to hopefully trigger the builds.

hainm:
ha, new failure is mpi test.

Issue #220, opened 2016-01-07
remove readline + libsander in libcpptraj


hainm:
I am testing pytraj in amber and got undefined symbol that relates to readline. Can  you remove it from libcpptraj since there is no use here. thanks.

> libcpptraj.so: undefined symbol: rl_attempted_completion_function

swails:
Why not just use the `-noreadline` switch when building libcpptraj for pytraj?

hainm:
I used the config.h file in $AMBERHOME

```bash
cd $AMBERHOME/AmberTools/src/
make pytraj
```

This is a part of Makefile
```bash
pytraj::
    (cd cpptraj && $(MAKE) -f Makefile_at libcpptraj)
```

so I am not sure where and when I can use `-noreadline` :(

rmcgibbo:
I tried building `libcpptraj.so` with just `./configure -shared; make libcpptraj`, and that symbol is definitely defined, according to `nm`. Could be a difference between the two Makefiles.

hainm:
@rmcgibbo: My error only appear when cpptraj is in Amber suite.

I think my problem is I did './configure -openmp gnu` first and then `make pytraj`. So libreadline is never built. :D (Installing Amber need `make install` for serial build first (@Jason: correct me if I am wrong) and do the parallel build later).

But removing libreadline from libcpptraj.so is still my wish since there is not use for it.

hainm:
@swails we discussed about checking serial build a while ago in email. May be it's time to implement.

(`Fwd: [AMBER] Problems with the installation of ambertools15.` (on Oct 2015)

hainm:
@Dan and Jason: can you double check if we need to make install serial version before openmp for cpptraj in amber? 

First I tried ./configure -openmp gnu

(Without serial build) then 'make cpptraj' but also got readline error. So readline is required in seriL built (which is quite complicated for cpptraj (vs standalone version)

drroe:
> @drroe and Jason: can you double check if we need to make install serial version before openmp for cpptraj in amber?

Just downloaded a fresh copy of the Amber GIT tree; `./configure -openmp gnu` followed by `cd AmberTools/src && make cpptraj` worked for me.

drroe:
> I think my problem is I did ./configure -openmp gnu first and then make pytraj. So libreadline is never built.

[...]

PR #336, opened 2016-05-25
New approach to building libcpptraj

drroe:
This PR represents a new approach to building libcpptraj. Instead of having to determine the pieces that will go into or be excluded from libcpptraj at configure time (e.g. having to specify `-nosanderlib` etc), make it a build time decision. This is done by having files that are different for cpptraj/libcpptraj be compiled into separate objects for libcpptraj using a libcpptraj-specific define, `-DLIBCPPTRAJ`. These objects have a `.LIBCPPTRAJ.o` extension (instead of just `.o`); this is a similar strategy to how SANDER compiles for LES.

The advantage of this is that no matter how cpptraj is configured, libcpptraj can still be built, so e.g. cpptraj can have sanderlib support, but the build does not have to be reconfigured to compile libcpptraj. Having separate objects for cpptraj/libcpptraj also means that parallel builds with both targets active will work.

This should hopefully address #220.


drroe:
@hainm I did  some testing with pytraj and it seemed to build fine, but I will be interested to hear your opinion. Ostensibly this should make it easier to make a trimmed-down libcpptraj in the future.

hainm:
thanks. I will test tonight.

hainm:
can you also merge this PR to amber repo too?

drroe:
Done.

hainm:
thx

Issue #311, opened 2016-05-02
make test.cuda


hainm:
related to https://github.com/Amber-MD/cpptraj/issues/283

drroe:
Addressed by #337

Issue #283, opened 2016-03-05
make test.openmp


hainm:
place holder to remind us to make openmp for amber. 

```python
amber (master) $ make test.openmp
make: *** No rule to make target `test.openmp'.  Stop.
```

drroe:
Addressed by #337

PR #337, opened 2016-05-25
Add more test targets.

drroe:
Minor, but does fix up binary determination now that multiple suffices are possible (e.g. OMP.MPI etc). Addresses #283 and #311.


drroe:
Since Travis passed and this fixes a problem with libcpptraj (and this shouldn't affect the windows build ::knocks on wood::) I will merge now.

PR #346, opened 2016-06-09
Major change to how pairwise distance matrices are handled

drroe:
This is a major change to how pairwise distance matrices are handled in clustering. Lots of functionality that was in the ClsuterMatrix class has been split among several classes. The pairwise distance matrix is now part of the data set/data file framework. Among other things, this allows multiple clustering commands to use the same pairwise matrix within the same run (so you can e.g. test different epsilons or what have you). In addition to the pairwise matrix stored within memory, there are 2 new options: no memory, i.e. distances are calculated on the fly (which will be slower but uses no memory) and disk, which caches the pairwise matrix on the disk (this option does not play well with OpenMP and so is not documented at the moment). There is also a new pairwise matrix file format introduced which uses NetCDF (this is now the default). Because it uses NetCDF, the new format is self-documenting and orthogonal, so can be expanded in the future with no issues. It also carries a description of the distance metric used to calculate it which can be useful.

Each individual ClusterDist::PairwiseDist() function has been removed; this functionality has been c
entralized to ClusterList::CalcFrameDistances(). Its a hair slower to do it this way (a few percent) but makes it so that if new distance metrics are added they dont need to implement PairwiseDist, and any improvements made to the pairwise distance calc fix all metrics.

Instead of using one class for frame-frame (i.e. pairwise) and cluster-cluster distances as was done
 previously, they are now separate (DataSet_Cmatrix and ClusterMatrix respectively). The functionali
ty has become different enough to warrant this and makes the code cleaner.

This PR also includes some fixes to the DBSCAN algorithm that make it somewhat more robust (and use less memory). Results should match previous ones to within a few percent. Also, some NetCDF function
ality that ended up being common to NetcdfFile and NC_Cmatrix was moved to its own namespace (NC) in NC_Routines.

@hainm I was able to compile pytraj with this OK. However, all tests fail with this kind of error:
```
======================================================================
ERROR: Failure: ImportError (cannot import name options)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/usr/lib/python2.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/usr/lib/python2.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/home/droe/GitHub/pytraj/tests/test_write_traj.py", line 4, in <module>
    import pytraj as pt
  File "/home/droe/Amber/GIT/amber/lib/python2.7/site-packages/pytraj-1.0.4b0-py2.7-linux-x86_64.egg/pytraj/__init__.py", line 61, in <module>
    from . import options
ImportError: cannot import name options
```
However, this happens with my master branch as well, so I believe it's unrelated to the changes in this PR.


hainm:
it's likely that you're using very old `pytraj`

`/home/droe/Amber/GIT/amber/lib/python2.7/site-packages/pytraj-1.0.4b0-py2.7-linux-x86_64.egg/pytraj/__init__.py`

pytraj version is `1.0.8.dev` now. :D 

I am not sure how you installed `pytraj`, but I think above is due to you `source amber.sh`, then `python` pick up the old `pytraj` in PYTHONPATH you set.

`/home/droe/GitHub/pytraj/`

```bash
unset PYTHONPATH
././run_tests.py
```

this should do the trick.

hainm:
just test `pytraj` and there is only 1 test failed related to clustering. I will update `pytraj`.

Can you update `cpptraj` version if you have not done that? need to force pytraj to use this PR.

drroe:
> Can you update cpptraj version if you have not done that?

Good call. I'll get to that soon.

PR #347, opened 2016-06-09
Add experimental command 'viewrst'

drroe:
Adds the `viewrst` command, but keep it hidden for now. This allows one to view Amber distance restraints in a visualization program (e.g. VMD) via construction of a pseudo-topology that uses bonds to illustrate the restraints. Also adds this functionality to the also experimental `nmrrst` command.


hainm:
should `viewrst` be another name? (I though it was viewing restart file).

drroe:
> should viewrst be another name? (I though it was viewing restart file).

Maybe eventually, not a big deal right now since it's a "hidden" command. I chose 'rst' since the 'rst' command in cpptraj has to do with restraints (also the 'nmrrst' command, sort of).

swails:
Or maybe rst2bond or something.

PR #350, opened 2016-06-09
CUDA support for all kernels

amit56r:
I have added the cuda support for all imaging kernel  (ACTION CLOSEST). They have been tested on BW system and they seems to be working.

### DIFF
--- a/src/Action_Closest.cpp
+++ b/src/Action_Closest.cpp
@@ -12,8 +12,8 @@
 #  include <cuda.h>
 
 // CUDA kernels
-extern void Action_NoImage_Center(double*,double*,double[3],double,int,int,float&,ImagingType,double[3],double[9],double[9]);
-extern void Action_NoImage_no_Center(double*,double*,double*,double,int,int,int,float&,ImagingType,double[3],double[9],double[9]);
+extern void Action_NoImage_Center(double*,double*,double[3],double,int,int,float&,ImagingType,double*,double*,double*);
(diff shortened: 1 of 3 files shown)


src/Action_Closest.cpp
```diff
@@ -318,11 +317,25 @@ Action::RetType Action_Closest::DoAction(int frameNum, ActionFrame& frm) {
...
     }
   }
-  // TODO handle imaging
+  
+
+  //low overhead so we will do it anyway
+  Box frmBox = frm.Frm().BoxCrd();
+  Matrix_3x3 ucell, recip;
+  frm.Frm().BoxCrd().ToRecip(ucell, recip);
+  double box[3] = {frmBox[0], frmBox[1],frmBox[2]};
```
drroe: @amit56r Any particular reason you are copying box coordinates and unit cell/fractional matrices to temporary arrays instead of using the underlying data pointers (i.e. Box::boxPtr() and Matrix_3x3::Dptr())?

amit56r: Ah! I was looking for that. I was not aware that we had access to the underlying pointers. I will make the change.

amit56r:
@drroe  Okay the changes has been made. And the Action_closest test for cuda run/pass

drroe:
Cool, nice work Amit. We'll eventually want to get the code well documented and I think some of the function names should be changed to better reflect what they do, but that can happen in a future PR. Thanks!

PR #355, opened 2016-06-17
Add support for reading separate force trajectory

drroe:
This adds the `mdfrc` keyword to `trajin`, allowing a separate force trajectory to be read in (similar to `mdvel` for velocities, which is now documented in the internal help). Also adds a test.


hainm:
it seems that you don't have the test that force is in a single file.

The created `CrdFrcVel.nc` has coords, vel and force.

drroe:
> it seems that you don't have the test that force is in a single file.

I don't need to test `CrdFrcVel.nc` directly; if something there screws up, either `Vel.crd` or `Frc.crd` will be written incorrectly. Reading/writing of NetCDF coordinates is already tested in numerous other places.

hainm:
ah, I see.

So right now cpptraj only supports reading `netcdf` force?

drroe:
No. If forces are present in the frame cpptraj will write them out. That's why the `CrdFrcVel.nc` file gets created with forces; otherwise we couldn't extract forces from it with `mdfrc`.

PR #371, opened 2016-07-02
Fix libcpptraj build

drroe:
Adds XDR related stuff to libcpptraj build for standalone and AmberTools.


hainm:
just tested your `drroe:fix_libcpptraj` and works for pytraj now. thanks

```python
In [2]: pt.load(nglview.datafiles.XTC, nglview.datafiles.PDB)
Out[2]:
pytraj.Trajectory, 51 frames:
Size: 0.006323 (GB)
<Topology: 5547 atoms, 349 residues, 1 mols, PBC with box type = ortho>
```

Issue #345, opened 2016-05-29
adding CPPTRAJLDFLAGS and CPPTRAJCXXFLAGS


hainm:
hi, after spending my weekend to try to make osx/clang/miniconda work with pytraj/libcpptraj and failed due to messing up with other packages in AMBER (e.g. emil). I think it's much easier to pass only extra flags to libcpptraj build.

Please check my patch here. To test, you should have miniconda/anaconda python.

```bash
git apply patch_osx_clang.txt
./configure -macAccelerate clang
make install
# or just checking pytraj first
# (cd AmberTools/src && make pytraj -j8 && cd ../tests/ && make test.pytraj)
```

[patch_osx_clang.txt](https://github.com/Amber-MD/cpptraj/files/288610/patch_osx_clang.txt)

Since this only takes effect with osx/clang/miniconda(or anaconda), other configurations are not affected.

drroe:
Should be addressed by #379

hainm:
Thanks. I will try.

hainm:
worked for me. thanks. close now.

PR #379, opened 2016-07-29
Fix binning of first torsion value in 'stat' analysis

drroe:
Related to #378. 

@hainm I also finally got around to addressing #345 (changes still need to be made to Amber configure though).


hainm:
> changes still need to be made to Amber configure though).

I just made this change and seems work for me.

```bash
./configure -macAccelerate clang
make install
```

Issue #68, opened 2015-09-16
Action_Density, write to DataSets


hainm:
I am working on this Action to write data to Dataset. Any suggestion which Dataset I should use?

(it's ok if saying `figure it out yourself`).

drroe:
Seems like this might be OK to use DataSet_double since the X values will always be monotonic. You will just have to make sure you set the X dimension for each set correctly, e.g.
```C++
Dimension Xdim( ((double)plus_minidx + 0.5) * delta, delta, plus_maxidx - plus_minidx + 1, "Density" );
data_->SetDim(Dimension::X, Xdim);
```
You should double check my math on the minimum and # bins but you get the idea.

drroe:
Addressed via #381

PR #381, opened 2016-08-03
Make 'density' command use data set/data file framework

drroe:
Also fixes the X axis label for sets output from `integrate`. Addresses #68 and #93.


hainm:
@drroe trick: if you just want to close an open issue (without writing more info), you can write "close #68" in your PR. The issue will be automatically closed when the PR is merged. :D

Issue #36, opened 2015-08-05
Split off general dihedral rotation functionality from Action_MakeStructure


drroe:
Create new action like 'rotatedihedral' with simplified syntax for rotating dihedral angles.

hainm:
this action would save lots of time and we don't need to write a tons of temp output files from tleap.

http://ambermd.org/tutorials/advanced/tutorial23/paramfit_1.html

>This script creates a random structure set using leap and saves each structure individually in a created directory. Then, it uses cpptraj to collate the structures together into a coordinate file.

drroe:
dihedralscan random could be adapted to do this (or maybe it already can?)

hainm:
I have not played with `dihedralscan` yet (since I don't know how to use it in `pytraj` + I prefer to see example first).

I just look at manual, it's likely an action to create random structure (which is really cool to me).

drroe:
From my extended test suite, 2 examples:
```
trajin ../tz2.rst7
# Rotate all phi psi dihedrals in intervals of 120 deg.
dihedralscan interval -120 outtraj rotations.mdcrd phi psi
```
And
```
trajin ../tz2.rst7
# Rotate dihedrals randomly; maxfactor relates to how many tries to attempt
dihedralscan random rseed 1 check maxfactor 10 phi psi
trajout random.mol2
```
Again, not sure how robust the code is, it's been a while since I used it. The manual has explanation of the keywords (I think).

hainm:
Thanks. I will try with pytraj.

On Wednesday, August 5, 2015, Daniel R. Roe <notifications@github.com>
wrote:

> From my extended test suite, 2 examples:
>
> trajin ../tz2.rst7
> # Rotate all phi psi dihedrals in intervals of 120 deg.
> dihedralscan interval -120 outtraj rotations.mdcrd phi psi
>
> And
>
> trajin ../tz2.rst7
> # Rotate dihedrals randomly; maxfactor relates to how many tries to attempt
> dihedralscan random rseed 1 check maxfactor 10 phi psi
> trajout random.mol2
>
> Again, not sure how robust the code is, it's been a while since I used it.
> The manual has explanation of the keywords (I think).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Amber-MD/cpptraj/issues/36#issuecomment-128173227>.
>


-- 
Hai Nguyen

hainm:
@drroe  

[...]

PR #382, opened 2016-08-03
Add 'rotatedihedral' command

drroe:
Used to rotate a specified dihedral to a specific value or by a given increment. Intended to be an easier to use alternative to `makestructure` for single dihedrals, as well as LEaP's `impose` command. Addresses #36.
```
> help rotatedihedral 
	crdset <COORDS set> [frame <#>] [name <output set name>]
	{value <value> | increment <increment>}
	{ <mask1> <mask2> <mask3> <mask4> |
	  res <#> type <dih type> }
	<dih type> = alpha beta gamma delta epsilon zeta nu1 nu2 h1p c2p chin phi psi chip omega
  Rotate specified dihedral to specified value or by given increment.
```


hainm:
may be adding new `label` since `rotatedihedral` is not an cpptraj's `Action`.

drroe:
> may be adding new label since rotatedihedral is not an cpptraj's Action?

Eh, there are already too many labels IMO :-). Would rather change 'new action' to 'new command' but not sure if that's possible. Anyway it just comes down to semantics, so I'm fine leaving the label as-is.

Issue #384, opened 2016-08-04
AtomMask::InvertMask code change?


hainm:
it seems that this 363515c80c1ac7bd754ec9f94d005b93455c66f2 commit broke `pytraj` test.

```python
        atm = AtomMask("@CA")
        assert atm.n_atoms == 0
        top = pt.load_topology("./data/Tc5b.top") # 304 atoms
        top.set_integer_mask(atm)
        assert atm.n_atoms == 20

        top2 = top._modify_state_by_mask(atm)
        assert top2.n_atoms == 20

        atm.invert_mask()
        top3 = top._modify_state_by_mask(atm)
        assert top3.n_atoms == top.n_atoms - 20
```

basically I am expecting to get `top3` with `304 - 20 = 284` atoms. However, I got `275` atoms.

Is this your intention?

drroe:
Nope, bug - I'll fix it.

drroe:
Can you see if it is fixed by #385?

hainm:
yeap.  thanks.

PR #385, opened 2016-08-04
Fix issue with inverting integer atom mask

drroe:
Should address #384.


hainm:
can you add a test for it?

PR #405, opened 2016-10-21
Comment out DEBUG statement 

drroe:
Thanks to @dacase for pointing this out.


drroe:
CI testing does not seem to have been triggered. However, since all I have done is comment out a DEBUG statement I think this should be good to merge (famous last words right?)...

PR #421, opened 2016-11-20
Improvements to Travis to help prevent possible false positives

swails:
Be more conservative about when to return an error (set -e means if any command results in a non-zero exit status, it kills the shell and returns non-zero).

Also work around this stupid thing on Mac Travis builds:

```
Done. Your build exited with 0.
/Users/travis/build.sh: line 148: shell_session_update: command not found
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,12 +27,14 @@ matrix:
     - { os: linux, env: COMPILER=gnu BUILD_FLAGS="-openmp -shared" OPT=openmp OMP_NUM_THREADS=1 BUILD_TYPE=libcpptraj TEST_TYPE=test.libcpptraj }
 
 before_install:
+  - set -e
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       brew tap homebrew/science;
       brew update;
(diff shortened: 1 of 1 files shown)


drroe:
Merging since I fully trust you; just curious what prompted the PR? Does the Mac Travis build end up strange sometimes?

swails:
Back in Jan. I found that some of the tests were failing but my test script wasn't picking up on it.  The reason for this is that the failing test did not occur on the last command, so the exit status was still 0.  While I *could* have just added a `|| error` to the end of every command and defined error like so:

```sh
error() {
    echo "An error happened"
    exit 1
}
```

you can just use `set -e` to effectively do that for you automatically.  The end result is that Travis stopped hiding test failures from me.

But in one of the recent Mac updates, one of Travis's scripts that they run as part of their test infrastructure was not updated and *always* tries to execute a failing command.  If you have `set -e` set, this failed command (trying to execute a non-existent shell function `shell_session_update`) then fails the entire stupid OS X test.

So to get around this, I just defined that function myself so the Travis build script has something to call.

PR #427, opened 2016-11-30
Add 'ucell' and 'shape' options to CHARMM Trajectory Read

drroe:
Add `ucell` and `shape` options to CHARMM DCD `trajin` to force reading of box info as unit cell parameters or shape matrix respectively.


drroe:
Mac/Clang build is what's holding up Travis. Everything else has passed so merging now.

hainm:
Dan: does this work with dataset trj yet? Is there any test for it?

drroe:
@hainm it should work with both `loadcrd` and `loadtrj`. There isn't a test because it's a weird case and I don't have any good input.

hainm:
ask `Brian Radak` for his input?

Issue #363, opened 2016-06-23
SPAM calculation not yet enabled


julimox:
Hello everybody,

I want to use SPAM and the calculation seems to run fine and also I get some output. However I doubt whether I can really trust it because during the calculations I get the message:

ACTION OUTPUT:
Warning: SPAM: SPAM calculation not yet enabled.

How can I enable SPAM?
Thank you very much!

drroe:
I'm going to ping @swails on this since he's the author of the SPAM code within cpptraj and can correct anything I say that is wrong.

My understanding is that the SPAM code in cpptraj is functional, but is slightly different from the original implementation in that it uses a simpler non-bonded calculation incorporating a force-/energy-shifted electrostatic calculation instead of full PME. The procedure is to first run the `volmap` command with the `peakfile` keyword in order to generate the density peaks needed by `spam`. Then you can run the `spam` action and post-process those results: see details in a post from @swails here: http://archive.ambermd.org/201401/0290.html

I think in principle SPAM could be fully implemented in cpptraj now since it has access to PME energies via the sander API, but that's a decent amount of work and I don't have a timeline for when it might be done.

julimox:
Thank you. I am using the python script he has uploaded in the same post and amber16. 
Let's see what he answers. Maybe there is a better way of using Spam now?

hainm:
> and I don't know how to install this chemistry module

there is no module name `chemistry` in AT16. It's renamed to `parmed`.

hainm:
So my suggestion is try to grep 'chemistry' and change it to 'parmed'. But let wait for Jason's feedback first.

swails:
@drroe has the situation pretty well described.  The energy evaluation in cpptraj is a simple energy-shifted nonbonded calculation, whereas the original implementation utilized 3 PME calculations within NAMD to give the energy of each "bound" water.  In the test systems published in the original SPAM paper, the results between the current cpptraj energy calculations and those by NAMD were pretty small (for example, one water position's energy distribution changed from ~-1 to -30 kcal/mol to +2 to 32 kcal/mol).  The averages were basically the same, but the energy shifting function gave a little more spread.

@hainm's suggestion for replacing `chemistry` with `parmed` and installing the ParmEd package (either through AmberTools 16 or the Github repository) are the same suggestions I would make.

drroe:
@swails so should the following warning be removed (or somehow modified)?
```
Warning: SPAM: SPAM calculation not yet enabled.
```

[...]

PR #429, opened 2016-12-01
Finish SPAM energy calculation, improve speed.

drroe:
This PR should make SPAM fully functional in cpptraj. Addresses #363. Notable additions:

- Calculations are OpenMP-parallelized and can now be run with MPI as well.
- The `summary` keyword works and will contain the free energy, enthalpy, and entropy values for each peak.
- Added keywords `[dgbulk <dgbulk>] [dhbulk <dhbulk>] [temperature <T>]` for specifying bulk free energy/enthalpy and temperature respectively; defaults to values for SPC/E water and 300.0 K.
- Code now uses the master data set list so energy values can be written in different formats etc.
- General code cleanup.

There is in general good agreement with this code and the old spamstats.py code. RMSE for free energy, enthalpy and -T * dS are 0.077491, 0.000011, and 0.077492 kcal/mol respectively.


swails:
Nice work!  This is awesome.

PR #431, opened 2016-12-05
Improve 'autoimage' for tightly packed systems

drroe:
This PR allows the `anchor` keyword for `autoimage` to use partial masks for centering, i.e. you can center on a specific region of a molecule instead of the entire molecule. Also adds a test which is an example where this functionality is necessary.


drroe:
Mac build on Travis holding everything up again. Since all other tests have passed I'm merging.

PR #432, opened 2016-12-08
Unhide 'comparetop' command

drroe:
Also add ability to compare atoms (not differing # atoms yet unfortunately) and a test.


hainm:
I already "unhide" it on `pytraj` :))

PR #436, opened 2016-12-12
Using configure for win build (appveyor) + update libcpptraj.dll.a

hainm:
Why `libcpptraj.dll.a`?
- to be consistent with recent change in AMBER
- help pytraj (`libcpptraj.lib` is not recognized)

Also using `configure ...` script rather the hard-coded `config.h`.


drroe:
Looks good.

Issue #417, opened 2016-11-12
missing tests


hainm:
- [ ] TRR and XTC formats
- [x] Action_Dipole - Its test is actually in `test/Test_Grid/`
- [x] Action_Channel - **updated** I see this Action is hidden. So never mind. :D 

(will update more)

drroe:
>  Action_Dipole - Its test is actually in test/Test_Grid/

This command is pretty much obsolete; it writes out in a format that is only readable by a plugin for the now-defunct MIDAS program. It's possible that the functionality has survived in chimera somewhere but I'm not sure. Bottom line is that I think it's fine to have it's regression test mixed in with `grid` since they share a good amount of code between them.

drroe:
Addressed by #446

PR #446, opened 2016-12-29
Fixes for AmberTools

drroe:
- Do not write header to output energy files in Test_CombineCrd since `dacdif` cannot use the `-I` flag.
- Remove `README` link to `README.md`; this has been removed in AmberTools and is no longer necessary.
- Add missing variable initializations to sander API energy call to avoid segfaults with `esander`.

Also add Gromacs TRR/XTC trajectory tests to address #417.


drroe:
@hainm The travis builds are failing:
```
Energy_Sander.cpp: In member function ‘void Energy_Sander::SetDefaultInput()’:
Energy_Sander.cpp:128:10: error: ‘struct sander_input’ has no member named ‘ibelly’
Energy_Sander.cpp:131:10: error: ‘struct sander_input’ has no member named ‘bellymask’
Energy_Sander.cpp:132:10: error: ‘struct sander_input’ has no member named ‘refc’
```
Seems that the Anaconda pysander is not up to date. In the meantime I'll have to improve the check for the sander API in cpptraj's `configure`.

drroe:
This is really frustrating. I don't like that cpptraj tests depend on a package which may or may not be out of date. I am going to disable sander API testing on Travis and rely on my own testing until a better solution is found.

hainm:
@drroe 

> Energy_Sander.cpp:128:10: error: ‘struct sander_input’ has no member named ‘ibelly’
Energy_Sander.cpp:131:10: error: ‘struct sander_input’ has no member named ‘bellymask’
Energy_Sander.cpp:132:10: error: ‘struct sander_input’ has no member named ‘refc’

I added those variables for sander (and pysander) in amber master. AmberTools16 does not have them.
You can still use current pysander distribution and exclude checking those variables. After AT17 is released, I will upload new version to Anaconda.

hainm:
uhm, you need to set default inputs. :D Do you want me to upload the development version of pysander?

drroe:
> uhm, you need to set default inputs. :D Do you want me to upload the development version of pysander (and sander)?

The problem is that when I did that the tests started failing because the version of pysander that gets pulled into travis _doesn't_ have those variables in `sander.h`. So the way things are this means that if something changes in the Amber GIT tree but doesn't change for the uploaded pysander and the sander API tests are active, my choices are either to have AmberTools tests fail or Travis tests fail. Neither of these appeal to me so I've chosen to disable the Travis sander API tests and do the testing on my own. I'm open to better solutions if they are out there.

hainm:
yeah, the best solution is having sander continuous integration so we can use the binary built. :D

PR #455, opened 2017-02-15
Improve the efficiency of the atomicfluct command

drroe:
Use an integer mask instead of character mask and only deal with the atoms that are selected.


drroe:
The Mac backlog is nuts...https://www.traviscistatus.com/

hainm:
I only test MacOS on my computer. :D Too long to wait for travis.

swails:
I can set up my Mac mini as a Jenkins slave and get OS X testing done there if you think it's worthwhile.

drroe:
@swails I may just take you up on that...

PR #461, opened 2017-02-28
Build fixes for OSX

drroe:
This PR fixes issues with clang++/gfortran linking on OSX. Also improves binary location logic.


hainm:
@drroe: can you turn on Pytraj compilation on macos too (jekin)? Should be quick to run.

Just need very simple test: 

import pytraj; pytraj.run_tests()

( few seconds to run). Thx.

PR #469, opened 2017-03-09
Add Ewald summation and direct sum methods to energy

drroe:
This is mostly just a test exercise to see if I could do it. It also potentially lays the foundation for adding native particle mesh Ewald in the future. Also adds a test and fixes an issue with `wavelet` segfaulting with OpenMP and FFTW.

Adds the following keywords to the `energy` command:
```
	[ etype {simple | directsum [npoints <N>] |
	         ewald [cut <cutoff>] [dsumtol <dtol>] [rsumtol <rtol>]
	               [ewcoeff <coeff>] [maxexp <max>] [skinnb <skinnb>]
	               [mlimits <X>,<Y>,<Z>]} ]
```

- `etype` chooses the type of electrostatics sum. Simple is what was previously there and the default.
- `direcsum [npoints <N>]`: Calculate electrostatics using the direct sum method for `<N>` cells in each direction.
- `ewald` is the Ewald summation method. Variables correspond to their **sander** equivalents.

It is actually faster than sander (mostly from speedups in the reciprocal sum), so if for some reason people need Ewald summation they can use **cpptraj** :-)


drroe:
Weird. Appveyor shows build failed but on the site it passed: https://ci.appveyor.com/project/drroe/cpptraj-aof9y
I guess I will try triggering a rebuild.

drroe:
GtG

PR #471, opened 2017-03-16
Fix for renamed docker image

drroe:
From @hainm


hainm:
sorry Dan, there's still one missing: 

https://github.com/Amber-MD/cpptraj/blob/master/util/test_pytraj_build.sh#L4

PR #473, opened 2017-03-17
Fix C++ library version issue of libcpptraj for pytraj

drroe:
When configuring for clang/macAccelerate on OSX make sure that the system c++ library is used. This fixes the issue where libcpptraj would be compiled with one version of the c++ library and pytraj would be compiled with another, leading to segfaults and other types of crashes.


hainm:
should you turn off appveyor? it's very annoying to see red 'x' :P.

PR #475, opened 2017-03-24
Fix for pytraj

drroe:
Make `PDBWRITEMODE` public again (done in #474) to fix pytraj build, and add a note so I don't forget.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: cpp
+dist: trusty
 sudo: false
 addons:
   apt:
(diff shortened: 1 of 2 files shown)


drroe:
This is frustrating - there's an SSL error with anaconda for pysander that is causing everything to fail.

hainm:
that's why I told @dacase that we should host our own binary distribution in ambermd.org server. (@Dave: from our discussion with John Chodera). Anaconda is very popular now so people tends to use it alot.

drroe:
Looks like switching to Trusty has fixed the issue, but for some reason the OpenMP build is just sitting in the queue.

drroe:
At long last...

Issue #392, opened 2016-08-26
Potential enhancements to 'hbond'


drroe:
- `nointrares` keyword?
- bridging time series?
- Ignore UA - V -UA bridges with nointramol hbond
- allow bridges within one residue
- hbond bridging track atom numbers

hainm:
@drroe 

per solvent bridge hbond ID:

`354(11+12+)` means solute 354 is bridge between res 11 and 12? is index 0-based or 1-based? 
what does the extra '+' in  `12+` mean?

drroe:
>  is index 0-based or 1-based?

1-based.

> what does the extra '+' in 12+ mean?

Nothing special - every residue number has a `+` after it.

envirodug:
@drroe, for the bridgebyatom new feature in hbond (thanks for this!), is it possible to add a few new features (or is it already possible to do this)?  

1. An option to only report heavy atoms in the output (e.g. instead of detailing hbonds to Lys HZ1, HZ2, and HZ3, just report to NZ).

2. Map the atom numbers to the residue number and name and print to the output file (e.g.  including 1:GLU, 2:ASP, etc. as in the hbond calculation without specifying bridgebyatom in the options).

drroe:
> An option to only report heavy atoms in the output (e.g. instead of detailing hbonds to Lys HZ1, HZ2, and HZ3, just report to NZ).

You mean in the average output right? An option doesn't exist for doing this yet - I'll put it on the to-do list.

> Map the atom numbers to the residue number and name and print to the output file (e.g. including 1:GLU, 2:ASP, etc. as in the hbond calculation without specifying bridgebyatom in the options).

Does the `printatomnum` keyword do what you want?

PR #485, opened 2017-04-18
Speed and functionality improvements for 'hbond' command

drroe:
This PR improves the single-threaded speed of `hbond` (particularly for solute-solvent hydrogen bonds), adds OpenMP support, fixes some minor bugs, and introduces several new options. This largely (maybe even completely) addresses #392.

The solute-solute hydrogen bond calculation speed has been improved by ~1.3x, and the solute-solvent calculation speed improved by > 3x. The single-threaded speed enhancement comes largely from refactoring the donor/acceptor loops. Previously there were a lot of duplicated distance calculations, particularly when it came to solute-solvent hydrogen bonds. Hints are now properly used in all maps which improves the find/insertion speed. OpenMP allows for even further speedup (have gotten 7.3x for 4 threads in a real-world test). 

Solute-solvent bridges within the same molecule are now properly ignored when `nointramol` is specified. Bridging interactions can now be tracked by solute atom, which allows better resolution when determining solute-solvent bridges; this is activated by the new `bridgebyatom` keyword (a test has been added for this). The sorting of bridges has also been made more stable. The code has been re-written to make it easy to add bridge time series data in the future.

Results will now be consistent for solute-solute hydrogen bonds in parallel (MPI and OpenMP) due to the more robust way solute-solute indices are calculated (no longer dependent on loop order). Also, `uuseries` and `uvseries` should now work properly in parallel. More detailed timing info can be turned on with the TIMER compiler define.

This PR also introduces the `sort` keyword for writing data files; this will ensure that data sets are sorted before being written out which allows consistent results between parallel and non-parallel versions of the code.

Version numbers have been incremented. We are now different from the official AT17 release.


drroe:
Travis seems slow today...

hainm:
@drroe: Because of this: https://travis-ci.org/Amber-MD/ambertools-ci
I let that repo watch Amber new commits (every hour), so ...

May be I will change to every 6 hours.

hainm:
only run nightly build for now since we just release AT.

drroe:
Cool, thanks for the info Hai. Everything is passed now so no worries.

PR #491, opened 2017-05-04
Add 'short' keyword to 'molinfo'

drroe:
Prints out molecule counts, e.g.
```
  [molinfo short]
#Mol Count Natom  Nres
DT5      1   939    30
DC5      1   963    30
Na+     58     1     1
WAT  13110     3     1
```


hainm:
what does `short` mean in this context?

drroe:
> what does short mean in this context?

Short form. For example, if I just type `molinfo` for the above example output I will get 13170 lines, one for each molecule. With `short` I get a more condensed view. Has similar spirit to `short` for `resinfo`.

PR #495, opened 2017-05-12
Enable the 'ti' analysis command and misc. other improvements

drroe:
Previously hidden since it wasn't really finished. I think it's ready for prime time. A test has been added. See below for an example of usage.
```
> help ti
	<dset0> [<dset1> ...] {nq <n quad pts> | xvals <x values>}
	[name <set name>] [out <file>] [curveout <ti curve file>]
	[nskip <#s to skip>]
	[avgincrement <#> [avgmax <#>] [avgskip <#>]]
	[bs_samples <samples> [bs_points <points>] [bs_seed <#>]
	 [bs_fac <factor>]]
```
Can be used to calculate free energy by integrating over DV/DL. An average DV/DL will be determined for each DV/DL data set, which will be used to create the curve that will be integrated. It is assumed that the DV/DL data sets are given in order of increasing lambda value.
- **nq \<n quad pts\>**: Integration will be done via Gaussian quadrature using specified number of points.
- **xvals \<x values\>**: Integration will be done via the trapezoid rule using the specified lambda values.
- **name \<set name\>**: Output data set name.
- **out \<file\>**: File to write integration result(s) to.
- **curveout \<ti curve file\>**: File to write raw DV/DL curve(s) to.
- **nskip \<#s to skip\>**: Calculate average DV/DL by skipping over specified number of points. \<#s to skip\> can be a comma-separated list; curves will be calculated for each skip value.
- **avgincrement \<#\> [avgmax \<#\>] [avgskip \<#\>]**: Calculate average DV/DL for increasing numbers of points, starting at point specified by \<avgskip\> (default 0) and ending at point specified by \<avgmax\> (default last point). So e.g. if avgskip is 50 and avgincrement is 100 average DV/DL will be calculated for 50-149, 150-249, etc.
- **bs_samples \<samples\> [bs_points \<points\>] [bs_seed \<#\>] [bs_fac \<factor\>]**: Calculate average DV/DL in "bootstrap" fashion; \<samples\> resamples of size \<bs_points\> points (default total number of points times \<factor\>) will be randomly selected and integrated. The average of all resamples (and standard deviation) will be reported. The random seed generated can be specified; if it is not the wallclock time will be used as a seed.

Data sets created:
- \<set name\> Results of integration.
- \<set name\>[SD] Standard deviation of bootstrap result (bootstrap only).
- \<set name\>[TIcurve][:\<X\>] DV/DL curve. For increment, the index \<X\> is the point number, for skip it is the number of points skipped, and for bootstrap it is the resample number (starting from 0),

This PR also has misc. related functionality.
- New keyword for `readdata`, `separate`. Can be used to specify that each file listed on a `readdata` command line should be read into separate data sets with `_<X>` appended (where X starts from 0) instead of concatenating.
-New keywords for `calc`: [prec <width>.<precision>] [format {double|general|scientific}]. The `prec` keyword controls output width and precision, while the `format` keyword can be used to control the format. E.g.
```
> calc 2*PI prec 16.8 format scientific
Result:   6.28318531E+00
```

TI Analysis Example: The following cpptraj input reads the output from 3 Amber MD TI simulations and performs both the "bootstrap" `ti` analysis as well as "incremental" to assess convergence of the results. A Gaussian quadrature of 9 points is being used. This assumes that the run directories are named Lambda_<X>, where <X> is constructed so the lambda values are in ascending order, e.g. Lambda_0.01592, Lambda_0.08198, etc.
```
readdata Lambda_*/run0.out name TI separate
readdata Lambda_*/run1.out name TI separate
readdata Lambda_*/run2.out name TI separate
ti *[DVDL] nq 9 name Curve out ti.dat curveout curve.agr \
  bs_samples 100
ti *[DVDL] nq 9 name Incr out incr.dat \
  avgincrement 500 avgskip 1000
runanalysis
```


hainm:
can `ti` be replaced by longer name?

drroe:
> can ti be replaced by longer name?

Any particular reason why? I think `ti` is a nice and concise name. I chose this because in the literature "thermodynamic integration" is quite commonly (you might even say universally) referred to as "ti". I don't think it's ambiguous but maybe there is something I'm not thinking of.

hainm:
ha, I was thinking about pytaj syntax, just don't look nice. :D 

```python
pt.ti(...)
```

But  you're right, We should keep `ti`.

PR #498, opened 2017-06-15
Make TI 'increment' analysis more robust to DV/DL sets of different sizes

drroe:
Useful when e.g. an individual lambda run (or runs) crashes.


drroe:
Seems like the failure is the GPU being down/unavailable on the Jenkins build bot, so I'm merging.

Issue #500, opened 2017-06-23
Intel 2018 Compilers OpenMP


midnight2103:
Intel 2018 compilers the -openmp flag is deprecated.  If possible switch to -qopenmp which is backwards compatible from 2016 onwards.

drroe:
Addressed by #502 - thanks for the report!

PR #502, opened 2017-06-23
Use '-qopenmp' flag for Intel compilers V16 and greater.

drroe:
Also better cleanup of files left behind by configure. Addresses #500.

### DIFF
--- a/configure
+++ b/configure
@@ -114,7 +114,7 @@ int main() {
 }
 #endif
 EOF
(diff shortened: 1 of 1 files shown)


hainm:
Can we turn off windows build? it's always failed.

drroe:
@hainm Yeah I don't know why the appveyor build hangs every now and then. It's usually fine if I restart it.

PR #504, opened 2017-06-27
New command for fixing bonds which have been distorted by imaging

drroe:
```
fiximagedbonds 
	[<mask>]
  Attempt to fix bonds which have been distorted due to imaging artifacts.
```
The `unwrap` command only works if the given reference structure is close enough to the first frame that atoms haven't moved too much. This command doesn't rely on a reference - instead, it looks for distorted bonds (longer than half the box size) which indicates an imaging artifact, then tries to correct the artifact. Works for both orthogonal and non-orthogonal cells. Adds tests for both cases.

Note that the command only tries to keep molecules together - it doesn't worry about imaging. Therefore it is recommended to run an `autoimage` after `fiximagedbonds`.


drroe:
Here the appveyor build shows as failed, but if I check the site (https://ci.appveyor.com/project/drroe/cpptraj-aof9y/build/1.0.551) it appears to be successful. I'll try re-triggering the build to see if that helps. If it doesn't work I'll make appveyor not required anymore.

drroe:
OK, I'm officially tired of appveyor hanging.

PR #507, opened 2017-07-03
Add 'guessbp' option to 'nastruct' to determine base pairing based on strand layout

drroe:
Adds the `guessbp` option to cpptraj `nastruct` which tries to guess base pairing in a very naive way based on how NA strands are laid out. It assumes that consecutive pairs of strands should base pair (i.e. strand 0 to strand 1, strand 2 to strand 3, etc) and that strands are laid out 5' to 3'. The `bptype` argument can be used one or more types to specify the pairing type between strands: `anti` or `para` (parallel). Can be useful for cases where it can be challenging to obtain base pairing via geometric criteria.


hainm:
why no test for this?

Issue #508, opened 2017-07-03
Add option to nastruct for specifying base pairing


drroe:
Similar to the recently added `guessbp` option (#507) but give users a bit more flexibility. Something like:
```
nastruct basepair :995-1141 :1142-1288
```
The `basepair` arg could be specified multiple times. Could also make use of the `bptype` args for this.

NOTE: I think it's safe to assume all NA strands are laid out 5' to 3' but could think about adding some sort of prefix to indicate the ordering is flipped, something like:
```
nastruct basepair :995-1141 threeprime :1142-1288
```

PR #510, opened 2017-07-05
New options for 'distance' and 'dataset', new tests

drroe:
Adds two new options to `distance`:
- Can calculate distance to a point in a reference structure.
- Can calculate distance to a specified point.

Add new option to `dataset`:
- **vectorcoord {X|Y|Z} [name <name>]**: Extract X, Y, or Z component of vector data set into a new data set.

Add tests for the above. Also add a test for `nastruct guessbp` (#508) for @hainm.


hainm:
>  Also add a test for nastruct guessbp (#508) for @hainm.

:D

hainm:
yeah, please. That's make easier to test pytraj.

PR #511, opened 2017-07-06
Add density calculation to 'volume'

drroe:
Since we're halfway there anyway in volume. Can now specify `density` keyword for `volume` action to get system density. Also makes it clear what units the data are in either case. Adds tests for both.


drroe:
Appveyor false positive. Merging.

hainm:
I have hard time with density command and density in volume command.

drroe:
Unfortunately `density` got scooped up by Hans Loeffler; his command does density along specific coordinates, not the entire system density. Instead of trying to force a weird command name I folded it into volume since they are related anyway. I guess I could either try to fold it in to the existing `density`  command somehow, or call it something like `systemdensity`? I don't like any of the options, including the one I went with.

Ideally I'd rename `density` to `coordinatedensity` and use `density` but then backwards compat. is broken...

hainm:
> Ideally I'd rename density to coordinatedensity and use density but then backwards compat. is broken...

Can we make SUPER NOISY warning in AT18 and make change in AT19? sometimes we really need to do that.

drroe:
> Can we make SUPER NOISY warning in AT18 and make change in AT19? sometimes we really need to do that.

I think a better alternative will be for me to fold it into the existing `density` command somehow. I'll look into this.

hainm:
Sounds a good plan. (I thought the same thing for pytraj).

Issue #512, opened 2017-07-07
Fold `volume density` into existing 'density' command


drroe:
Should hopefully be less confusing. Need to make sure I don't mess with existing syntax.

hainm:
If this happens, we might want to not advertise "volume density" syntax.

drroe:
Addressed by #513

PR #513, opened 2017-07-07
Enhancements and fixes for setvelocity, temperature, density/volume

drroe:
- Fix the degrees of freedom calculation used by `temperature` (had a -6 leftover from somewhere).
- New Constraint class for dealing with constraints etc.
- Modify the `setvelocity` command:
  - Add ability to specify constraints and use RATTLE algorithm to remove velocity components along constrained bonds.
  - Add ability to zero total momentum of atoms in mask.
  - Add ability to modify existing velocities instead of setting new ones.
- Frame::SetCoordinatesByMap() now correctly handles force info if present.
- New type of `vector`, `momentum`
- Total system density calc removed as keyword in `volume`, is now obtained via the `density` command when no masks are specified; addresses #512.
- Related help updated, and tests added/updated.


src/Action_SetVelocity.cpp
```diff
@@ -3,28 +3,67 @@
...
-  mprintf("\t[<mask>] [tempi <temperature>] [ig <random seed>]\n"
-          "  Set velocities in frame for atoms in <mask> using Maxwellian distribution\n" 
-          "  based on given temperature.\n");
+  mprintf("\t[<mask>] [{tempi <temperature> | modify}] [ig <random seed>]\n"
+          "\t[%s] [%s]\n", Constraints::constraintArgs, Constraints::rattleArgs);
+  mprintf("\t[zeromomentum]\n");
+  mprintf("  Set velocities in frame for atoms in <mask> using Maxwellian distribution\n" 
+          "  based on given temperature; default 300.0 K. If tempi is 0.0 set\n"
+          "  velocities of atoms in mask to 0.0.\n"
+          "  If 'modify' is specified do not set; only modify existing velocities.\n"
```
hainm: what is the different between 'set' and 'modify'?

swails: The way I read it is that velocities are only set from a M-B distribution if velocities were in the original frame if you use the `modify` keyword.  By contrast, without `modify`, velocities are set regardless.

drroe: > what is the different between 'set' and 'modify'?

The default mode `set` is for setting velocities from a Maxwell-Boltzmann distribution; this happens whether or not the incoming frame has velocity information or not. The `modify` keyword essentially means do not do this, just modify any existing velocities. This allows `setvelocity` to be used to e.g. zero out the total momentum in a system via `zeromomentum` or to modify velocities to be consistent with constraints via `ntc`.

PR #514, opened 2017-07-10
Try to debug random hangs on appveyor

drroe:
Every now and then, tests will just randomly hang on appveyor. Never at the same point, but always during the test phase. This PR is an attempt to reproduce and debug this Heisenbug...


drroe:
Hm...of course the appveyor build succeeds now. I will keep trying to rebuild until it fails or until I get tired.

drroe:
This time it [hung](https://ci.appveyor.com/project/drroe/cpptraj-aof9y/build/1.0.590#L2090) on executing cpptraj right after a previously successful execution:
```
Warning: CPPTRAJHOME is not set.
-rwxr-xr-x 1 appveyor None 35813182 Jul 12 12:16 ../..//bin/cpptraj
**************************************************************
TEST: /c/projects/cpptraj-aof9y/test/Test_Diffusion
  CPPTRAJ: Diffusion test, no imaging.
   ../..//bin/cpptraj  ../tz2.truncoct.parm7 diffusion.in >> test.out 2>>/dev/stderr
DEBUG: Cpptraj exited with status 0
DEBUG: DoTest noimage.dat.save noimage.dat
DEBUG: DoTest noimage.agr.save noimage.agr
  CPPTRAJ: Diffusion test, old syntax.
   ../..//bin/cpptraj  ../tz2.ortho.parm7 diffusion.in >> test.out 2>>/dev/stderr
```
This is starting to feel like it's just something that happens on appveyor. I may need to execute a smaller subset of tests. Or give up debugging this for now...

drroe:
The scope of this PR has expanded quite a bit now. This is now a complete overhaul of the test system. There are two major changes.

1. **Improved environment setup.** I had a feeling that one of the reasons the appveyor build was so prone to failure is that the old test system was not very smart about setting up the environment; it was done for _every single test_. This entailed calling cpptraj many times per test script, putting extra strain on whatever test platform. The MasterTest.sh script has now been overhauled so it is smart about how it is called. If called from above test directories it will perform environment setup and export certain variables. If called from a test directory it will only try to setup the environment if not already done. In terms of testing time it cuts down the full serial test suite timing quite a bit. On my sandybridge desktop total test time **went from about 69 s to about 36 s**.

2. **Better test environment detection and reporting of skipped tests.** The mechanism used by tests to determine if the cpptraj binary and environment are good for the test has been simplified from many individual functions to just two functions. For example, previously if a test depended on NetCDF, Zlib, and Bzlib one would have:
```
CheckNetcdf
CheckZlib
CheckBzlib
```
The problem is these were sledgehammers - they would just exit out with a message but not report it in any of the output stats. With the new framework now you would do:
```sh
TESTNAME='something'
Requires netcdf zlib bzlib
```
Not only is that more compact, but the test framework will report that the test was skipped in the final summary etc. Individual test units are skipped like so:
```sh
UNITNAME='part of a test'
CheckFor notparallel maxthreads 4
if [ $? -eq 0 ] ; then
  <Do the test>
```
This is also reported in the final output.

drroe:
Well, appveyor passed - here's hoping this PR did the trick and this is not a fluke.

hainm:
image

PR #516, opened 2017-07-13
CHARMM replica log/trajectory processing fixes etc

drroe:
- Automatically search for replica logs when `nrep` not specified.
- Fix sorting by coordinate with CHARMM replica log data.
- Fix output of CHARMM timestep when debugging (time variable still not used).

Also clean up and fix help text, and enable two previously skipped tests that should be run.


drroe:
Something strange happened with the Jenkins build - hiccup with the Mac box. Going to merge.

PR #522, opened 2017-07-28
New lipid order parameter calculation

drroe:
This PR introduces a new command for calculating lipid order parameters SCD, `lipidscd`.
```
	[<name>] [<mask>] [{x|y|z}] [out <file>] [p2]
  Calculate lipid order parameters SCD (|<P2>|) for lipid chains in mask
  <mask>. Lipid chains are identified by carboxyl groups, i.e.
  O-(C=O)-C1-...-CN, where C1 is the first carbon in the acyl chain
  and CN is the last. Order parameters will be determined for each
  hydrogen bonded to each carbon. If 'p2' is specified the raw <P2>
  values will be reported.
```
This is differentiated from the existing `lipidorder` command by the following:
- Automatic search for lipids; no need to specify individual carbon names. Unique lipid chains are identified by their carboxyl group atom and residue name. Saturated vs unsaturated carbons are handled automatically.
- Absolute value of <P2> is reported instead of -<P2>; conversations with Rich Pastor convinced me this is correct. There is an option to report the raw <P2> values as well.
- Full data set and data file framework integration. Two data sets are created for each chain, an average and standard deviation for each carbon.
- Works with OpenMP and MPI.
- No option to work with united atom; requires hydrogens present.

A test has been added for this command. This PR also has the following minor fixes/changes:

- Introduces the `xydy` option for Grace output, which allows automatic creation of XYDY data from consecutive data sets; useful e.g. for average/standard deviation data sets from `lipidscd`.
- Handles automatic creation of the SEGID field for PSF output in a much more sane fashion.
- Fix minor memory error when stripping Topology with no angle/dihedral parameters.


hainm:
What does SCD stands for? ￼

drroe:
> What does SCD stands for? 

SCD is the typical name given in the literature for the carbon-deuterium acyl chain order parameter determined from NMR. See e.g. [Vermeer et al., Eur. Biphys. J. (2007) 36:919-931](https://www.ncbi.nlm.nih.gov/labs/articles/17598103/)

hainm:
Thanks Dan.

hainm:
Are there many types of lipid oders?

hainm:
It took me a while to realize this

```
Command::AddCmd( new Action_LipidOrder(),    Cmd::ACT, 1, "lipidscd" );
```

Why not change `Action_LIpidOrder` to `Action_LipidSCD` like others?

swails:
+1 for matching class names with the verb used in the command interpreter.

drroe:
> Why not change Action_LIpidOrder to Action_LipidSCD like others?

Originally this started out as a drop-in replacement for `Action_OrderParameter` (whose command is `lipidorder`). In fact, during testing I was using `lipidorder2` as the command name. However, I haven't replicated all of the functionality (specifically estimation of C-D order parameters from only C positions) and the command has evolved to use significantly different input and produce slightly different output, so there wouldn't be any backwards compatibility. However the command is useful enough that I think it should be out there. So I settled on `lipidscd` being a nice compact command name. Renaming `Action_LipidOrder` just to match is very low on my priority list but maybe it will happen some day.

PR #524, opened 2017-08-01
Fix build with AppleClang in C++11 mode

multiplemonomials:
When set to the C++11 standard, AppleClang will not allow you to use forward declarations as template arguments to `std::map` (and others).  This is breaking the CMake build on macs.  I supposed I could set cpptraj to build in C++98 mode, but I feel like we should just fix the issue.  I moved the classes forward in the headers so that the forward declarations are not necessary.


drroe:
Why is Apple always so obnoxious?

Anyway thanks for doing this. Ultimately the forward declarations were just to make things look cleaner so it's not a big deal.

PR #529, opened 2017-08-15
Fix 'autoimage moveanchor' for non-orthorhombic cells

drroe:
Switching from Cartesian to fractional coordinate systems can "stretch" distances, so the approach used to determine the closest translation vector for orthorhombic cells with the `moveanchor` option did not always work for non-orthorhombic cells. Also simplifies the orthorhombic case - no longer need the pseudo `ANINT` call - just use `floor()`.


drroe:
Appveyor failure is not real.

PR #530, opened 2017-08-21
Rewrite pair list class; large speedup for 'checkstructure'

drroe:
This PR consists of two major changes/improvements to cpptraj:

1. Rewrite the PairList class to make it easier to use and incorporate into Actions etc.
2. Incorporate the PairList class into the `checkstructure` Action to speed up systems with box info, which tend to be large.

The PairList class was originally written for the Ewald class and was based on how SANDER employed a pair list. While it worked, it was not written in a very user-friendly or object-oriented way. It has now been written to be much more object oriented: the pair list consists of cells (`CellType`), to which atoms are added (`AtmType`). Cells contain information on all neighboring cells. The performance is unchanged from the previous implementation (based on tests of the `energy ewald` command) and it is now easier to iterate over cells and atoms in cells.

With the PairList class now being easier to use it has been incorporated into the `checkstructure` Action as a proof-of-concept and performs extremely well. In tests of some large systems (thousands to tens of thousands of atoms) the version of `checkstructure` with a pair list shows a speedup of over 100x compared to the non-pair list version.

The structure checking aspect of Action_CheckStructure has now been properly separated into the new `StructureCheck` class so that other parts of the code (e.g. `Exec_PermuteDihedrals`) get access to only the functionality they need. The `checkstructure` command now also works properly with MPI and gives consistent results when used with MPI and/or OpenMP (i.e. order of problems found is the same no matter what). The OpenMP version scales slightly better now due to the removal of `critical` pragmas.

This PR also adds more tests for `checkstructure`, including for the new pair list functionality.


drroe:
Jenkins failure looks like an issue with the build system - safe to merge.

PR #532, opened 2017-08-24
Add INSTALL_HEADERS option to CMake build system

multiplemonomials:
This make it a bit easier to use pytraj with the CMake build system.  When `INSTALL_HEADERS` is set to `TRUE`, then all of cpptraj's headers will get copied to the installation directory, and someone building pytraj can just set their `CPPTRAJ_HEADERDIR` variable to point there.

Also, I don't know if you followed [this PR](https://github.com/Amber-MD/cmake-buildscripts/issues/29), but the gist of it is that we need to add a few dllimports and dllexports to cpptraj to get it to work properly in Windows (also Pytraj works in Windows now, yay).  Probably this would mean putting something like the following in a central header:
```c
#ifdef _WIN32
# ifdef CPPTRAJ_EXPORTS
#  define EXPORTED_SYMBOL __declspec(dllexport)
# else
#  define EXPORTED_SYMBOL __declspec(dllimport)
# endif
#else
# define EXPORTED_SYMBOL
#endif
```
and then putting the `EXPORTED_SYMBOL` qualifier on every static class member.

This will make it so CMake 3.4+ can be used on Windows.  Apparently, the CMake people are working on fixing the auto-export of constants, but I don't know when or if it will be merged.  So we also have the option of waiting a few months to see if this becomes unnecessary.  However, if you don't mind putting these ugly declarations on your code, I'd prefer to fix the issue now.  Let me know whether you want to add the import declarations, and where to put the above block, and I can get started.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,15 +48,31 @@ if(NOT INSIDE_AMBER)
 	set(BUNDLE_IDENTIFIER org.ambermd.cpptraj)
 	set(BUNDLE_SIGNATURE CPTJ)
 	include(Packaging)
+	
+	# header installation option
+	option(INSTALL_HEADERS "Copy headers to the include/cpptraj folder of the install directory.  Useful for building with pytraj." FALSE)
+else()
+	set(INSTALL_HEADERS FALSE)
 endif()
 
 # on Windows, make MSVC auto-create import libraries just like MinGW does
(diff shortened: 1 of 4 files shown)


drroe:
Sorry for the delay. Crazy week this week with ACS and an additional workshop at NIST. WRT the dllimports stuff I'll need to review that. I'm not in a huge rush to get things working on windows - my sense is that there isn't a huge market for that with cpptraj but maybe I'm wrong about that.

PR #536, opened 2017-09-01
Add test for areaperlipid action

drroe:
Also output number of molecules being used during setup when `nmols` specified.


drroe:
Appveyor is hanging again...

hainm:
+1000 to kill it.

drroe:
Eh, it costs nothing beyond some annoyance to keep it and it's currently the only windows environment test. I'm working on ways to make the test framework a little less heavy than it is in the hopes it will ease things for appveyor (#514). We'll see...

Issue #537, opened 2017-09-03
test failed in AT


hainm:
Dan,

I've been getting cpptraj failures within AT (tried with my Mac, travis, circleci, ...)
Can you try that? thanks

```
==============================================================
===================== TEST SUMMARY ======================
  0 out of 1 comparisons OK (1 failed, 0 warnings).
  0 out of 0 tests completed with no issues.
=========================================================
make[2]: *** [summary] Error 1
make[2]: Leaving directory `/home/haichit/amber_git/amber/AmberTools/src/cpptraj/test'
make[1]: *** [test.all] Error 2
make[1]: Target `test' not remade because of errors.
make[1]: Leaving directory `/home/haichit/amber_git/amber/AmberTools/src/cpptraj/test'
make: *** [test.cpptraj] Error 2
```

hainm:
bonus failure (from circleci AmberTools test (created from its tarfile via mkrelease_at).

```
  CPPTRAJ: Data filter test.
Error: '../Test_Diffusion/diff_a.xmgr': No such file or directory
Error: Could not read data file '../Test_Diffusion/diff_a.xmgr'.
	1 errors encountered reading input.
Error: Error(s) occurred during execution.
Error: cpptraj exited with status 1
possible FAILURE:  file datafilter.dat does not exist.
==============================================================
**************************************************************
```

drroe:
Should be addressed by #538.

hainm:
thanks

PR #538, opened 2017-09-05
Use test save file instead of generated file.

drroe:
Removes dependency of Test_Filter on Test_Diffusion executing first. Should address issue #537.


hainm:
appveyor again.

Issue #540, opened 2017-09-09
add "fast" target for Makefile?


hainm:
Just for "make fast" in AmberTools. Ideally, "fast" should be < 10 s (ideally few seconds).

Current syntax is
```
cd some_folder && $(MAKE) fast
```

drroe:
I'll work on this on Monday.

FYI, cpptraj does have a 'test.short' target which covers a lot of functionality and only takes only 16 s on my desktop. However, because dacdif is so heavyweight the AmberTools tests take 46 s...

hainm:
On my mac, stand alone cpptraj "test.short" takes 43 s.

drroe:
I'm having second thoughts about this. On my Macbook Pro (2.5 GHz i7) `make test.all` takes 53 s (not including the `Test_Esander` test). I don't think that is unreasonable considering all the various functionality that is tested. I can definitely work on shaving time off of the more expensive tests like `Test_Esander`, `Test_Volmap`, `Test_GIST`, `Test_Ewald`, and `Test_SPAM`, but I think that 10 s is an unreasonable constriction considering all of the functionality that is in cpptraj:
```
DEBUG: 186 commands.
DEBUG:	               General : 25
DEBUG:	                System : 1
DEBUG:	                Coords : 7
DEBUG:	            Trajectory : 5
DEBUG:	              Topology : 17
DEBUG:	                Action : 77
DEBUG:	              Analysis : 35
```

drroe:
> On my mac, stand alone cpptraj "test.short" takes 43 s.

BTW this is because I forgot to update `make.short` to use the new test framework.

hainm:
Hi

I am talking about 10 s (x >500 tests in AT) in the context of 'make fast'
target, not for the full test suite.

The (almost) full test suite still run nightly.

Hai

drroe:
> I am talking about 10 s (x >500 tests in AT) in the context of 'make fast' target, not for the full test suite. 

[...]

PR #545, opened 2017-09-20
Complete test coverage of actions/analyses, many functional improvements

drroe:
This PR was originally started to evaluate and cut down on some of the more time-consuming tests, and has expanded to include that and increase test coverage, as well as many functional enhancements. Many of these were developed to address issues that popped up when added/evaluating the test suite.

The changes are (in no particular order):

- Every command that is not hidden and that does some kind of calculation (so not e.g. system commands like `ls` or `xmgrace`) now has a test. Added specific tests for: distance based masks, `calc`, `calcstate`, `change`, `createreservoir`, `createset`, `fixatomorder`, `lowestcurve`, `minimage`, `regress`, `meltcurve`, `multihist`, `rst`, `pairwise`, `parmbox`, `solvent`, `translate`, `replicatecell`, `scaledihedralk`, and `phipsi`.
- Very time consuming tests are now not run by default and have had shorter versions added. The longer versions can be run by specifying an option. Long tests are always on for CI testing. This pretty much addresses #540. Because of this, the `short` test option is now not needed.
- Grid and Matrix templates now use unsigned/long integers for indexing; should be able to handle very large systems now.
- The `bounds` command now stores X min/max etc in data sets.
- The temporary topology written by `Energy_Sander` is now removed on destruction unless `keepfiles` is specified.
- Add the `scalepdbe` option to `pairwise` to write energies in the output PDB scaled from 10 to 100 to avoid overflowing the bfac/occupancy columns. 0.0 means the energy was outside the cutoff. The `pairwise` action also provides more details on what it is doing and how it is set up. The `avgout` file now uses the master data file list.
- Reorganize help for `spam` so it is clear which options are for `purewater` and which are not. Also fixes an issue where energy was not calculated if `out` was not specified.
- Add `out` option for `divergence`.
- Add `nx` option for `regress` to create output data set with specified number of X values ranging from input set min X to max X. Also store slope and intercept in data sets.
- Add `createset` command to create a data set using a formula. Good synergy with e.g. `regress`.
- Make `invert` standard data write respect `xprec` option.
- Make distance based mask calc 20% faster. Also fix the behavior of the "outside" distance-based mask.
- Data sets created by `RPNcalc` class now have their formatting properly set.


drroe:
run jenkins

Issue #71, opened 2015-09-16
Expand mask syntax


drroe:
This is related to ParmEd/ParmEd#209.

I would like to expand the mask parser to be able to select molecules, chain IDs, etc. 
# Current proposals from @swails

A couple ideas:

Do what Chimera does (see https://www.cgl.ucsf.edu/chimera/docs/UsersGuide/midas/frameatom_spec.html)
Use :: for chain IDs (in the same way that @% indicates atom type names)
An example of 1:

```
# Select residues 1-10 in chain A
:1-10.A
# Select residues 1-10 in chains A, B, and C
:1-10.A-C # I *think*; not sure if you can use - for string range, maybe .A,B,C
# etc...
```

An example of 2:

```
# Select chain A
::A
# Select residues 1-10 of chains A, B, and C
::A,B,C:1-10
# Select protein backbone atoms of residues 2-20 in chains A and B
::A,B:2-20@CA,C,O,N
```

Another idea for flexibility is to say that chains are strings, and use the same "chain" syntax to denote molecules when the chain is given a numerical value.
# My Thoughts

I do like 2) for chain IDs - I think we could potentially get into trouble with the `.` syntax. I don't know why but for some reason I feel I've come across atom names with `.` in them, although nothing is coming to mind now.

For molecule selection, personally I like the idea of having a separate character that can represent molecules, the  way there is currently `@` for atoms and `:` for residues. Based on what is currently used in the mask parser, I think that the most likely candidates for "molecule" are `;` and `%`. Both have drawbacks: `;` looks enough like `:` that things could easily get confused (and it's on the same key), while `%` is already used in a subset of `@` to denote "type", so it would have different meanings in different contexts.

Whatever is decided, it should be consistent between ParmEd, Cpptraj, and the rest of Amber.

hainm:
wow, already 2 years. Time is running fast.

drroe:
Addressed by #548

PR #548, opened 2017-09-26
Expand mask syntax

drroe:
At long last, trying to address #71.

This PR expands cpptraj's mask selection syntax by introducing the following:
- Selection by molecule index. Uses the caret `^` character, e.g.
```
^1-2:ASP
```
would select residues named `ASP` in molecules 1 and 2. I decided on `^` because it's highly unlikely to be used in an atom/residue name and it had no overlap with existing operators/operands. @swails, @hainm, @dacase, can you think of any reasons why not to use `^` here?
- Distance-based selection by molecule. Use `^` in distance-based mask instead of `:` or `@`, e.g.
```
@5<^3.0
```
would select all molecules within 3.0 Angstroms of atom 5.
- Selection by chain ID. Uses forward slash `/` modifier for residue `:`, e.g.
```
:/B
```
would select all residues with chain ID `B`.
- Selection by original (typically PDB) residue number. Uses semicolon `;` modifier for residue `:`, e.g.
```
:;2-4,8
```
would select all residues originally numbered 2-4 and 8.

This also includes a slight change to how argument lists search for masks. Previously an argument could be considered a mask if it contained `@`, `:`, or `*` anywhere in the argument string. The new method only considers an argument string a mask if it contains `@ : ^ *` or `=` after any leading `!` and `(` characters. This is more robust, but can lead to situations where bad masks are ignored and effectively changed to `*`. For example, with previous method `select test@` fails with `Tokenize: Wrong syntax` because `test@` is treated as a mask, but with the new method `test@` is ignored and `select` assumes no mask present and selects everything. Not sure which behavior I like better and would be interested in opinions from @swails and @hainm.

This PR also includes some cleanup of the MaskToken and MaskTokenArray classes, and adds tests for the new syntax.


hainm:
@drroe let me have a bit thinking first.

hainm:
Just quick question. 

Will you update mask selection for other programs in AT too? then ParmEd needs to update too.

hainm:
Saying

- 
`^1-2:ASP@CA`: does this mean "select CA atoms from ASP in mol 1 and 2"? If yes, can you add test for that?

- so on for `@5<^3.0@CA`

Overall, I think the change is pretty good. Will add some testing for pytraj for sure.

swails:
I'll respond when I'm on my computer tonight.

swails:
> Selection by chain ID. Uses forward slash / modifier for residue :, e.g.

What Chimera does here is :: for chains, which is a scheme familiar to a non-negligible crowd of users and from their experience seems to have been working fine for quite some time.  Any reason not to do the same?  This could actually double as the molecule selector where integers select molecule numbers and letters select chains.  The arguments for this approach are

1. Chains and molecules are very similar conceptually and I've yet to see any chain named with a number so there won't be any practical conflict.
1. Other operators behave the same way in terms of treating integers as indexes and alphanums as names (@ and :)

The arguments against this approach are

1. Chains and molecules are similar, but also different.  Cross-linked chains are the same molecules, and using the same syntax for both may lead to confusion (and S-S bonds are quite common).
1. Solvent is typically all one chain, but are all separate molecules
1. You could also just have `::<number>` select the `<number>`th chain.

If the arguments against are more persuasive, I think ^ is a reasonable character to use for molecules. In fact, I think point 3 above has pushed me away from this idea anyway (and in support of chain #). However, I do like :: better than :/ for chain selection.  The former is what Chimera does and the latter will be rendered into an emoji whenever you try to send it to someone :smirk:

> Selection by original (typically PDB) residue number. Uses semicolon ; modifier for residue :, e.g.

What would happen if someone used this syntax on a prmtop file that wasn't modified via `addPDB`?  It would probably be quite perplexing to users that didn't realize prmtop files don't store the information needed to use `:;` by default (after all, the prmtop is a bit of a black-box format), so maybe an error is the way to go.  But I don't know.  I can see the utility of original residue number, but I don't have strong feelings about it at the moment.

As a general question, what happens when selecting based on information that doesn't exist? (Chain, original residue number, ... not sure if I'm missing one.)

drroe:
Thanks @hainm and @swails for the feedback! I'll address concerns in order.

> Will you update mask selection for other programs in AT too? then ParmEd needs to update too.

I will try to make this a priority, I don't have a lot of time for development these days. There are quite a few programs that would need to be updated: parmed, sander/ambmask, pmemd (feels like there should be a common parser for those two), mdgx, nab, and probably some I'm missing. I'd at the very least create an Issue or bug report asking for these to be added :-).

> ^1-2:ASP@CA: does this mean "select CA atoms from ASP in mol 1 and 2"? If yes, can you add test for that?

Yes, it does mean that, it does work, and I'll add a test.

>    so on for @5<^3.0@CA

There is already a test for distance selection by molecule in Test_DistBasedMask (unless I'm not understanding your meaning).

> What Chimera does here is :: for chains ... Any reason not to do the same?

I did try to implement this initially, and to make a long story short the way the cpptraj mask tokenizer currently works (character by character) this is not straightforward to do. I'd have to introduce some extra logic that says "if this character is `:` and the character _just prior_ was also `:` ..." (currently it just tracks "was `:` the last selection token I encountered?"). This admittedly isn't conceptually hard but is potentially time consuming - really best solution is probably to just rewrite the tokenizer. And as we all know rewrites often sound like they'll take 2 hours tops and then end up taking 2 days :-). That I'm willing to do the work if it's the best solution - get it right the first time.

> Chains and molecules are similar, but also different. 

I would go farther and say that they're very different. In a non-insignificant fraction of PDBs I have encountered, chains span multiple molecules (for example in dimers/tetramers etc all molecules that make up a monomer are often given the same chain ID).

> What would happen if someone used this syntax on a prmtop file that wasn't modified via addPDB?

For chain ID, you'd get no atoms selected. For original residue number, if it's not present cpptraj fills it in with 1-based ordering (so e.g. residue number and original residue number are exactly the same for Amber topology files).

To to summarize it seems that `^` is good for molecule selection, and `::` is recommended for chain ID selection. If the latter is implemented should `:/` be used for original residue number, or keep it as `:;`? The only reason I used `:/` in the first place was because `/` is already a "modifier" for `@`, but I'm not set on using it.

hainm:
>  @5<^3.0@CA

I mean "does this select CA atoms of molecule within 3 A from atom 5?"

drroe:
> I mean "does this select CA atoms of molecule within 3 A from atom 5?"

No, but to be fair it doesn't mean that or `@` or `:` either - you need to explicitly provide the `&`.

drroe:
Turns out the tokenizer modification wasn't as bad as I thought. Chain ID selection is now `::`.

drroe:
@swails @hainm any comments on the change in how argument lists search for masks (final paragraph of the original PR post)?

hainm:
No I have not.

drroe:
In that case I'll merge now - easy enough to change behavior back later if needed.

PR #552, opened 2017-09-28
CMake 1.1 changes

multiplemonomials:
This is the same changeset as last time, except now it won't contain unintended changes to other code.  Not sure what happened there, sorry about that.  

I ran the tests on the CMake-built binary, and they all seemed to pass.

By the way, I only now noticed that cpptraj's build system supports building combinations of the MPI, OpenMP, and CUDA versions.  I have updated the CMake build system to be able to build all combinations now, up to and including `cpptraj.MPI.OMP.cuda`.  Enjoy!

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
-[submodule "CMakeConfigScripts"]
+[submodule "cmake-buildscripts"]
 	path = cmake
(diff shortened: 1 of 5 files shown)


drroe:
This one seems good - thanks!

PR #556, opened 2017-10-04
Add control structures to cpptraj scripts

drroe:
This PR adds a control structure framework to cpptraj scripts; specifically `for` loops. Since it represents a significant feature addition I'm keeping it a work in progress for the time being.

The `for` loop can currently be over atom masks or integers. Help entry:
```
	{ {atoms|residues|molecules|molfirstres|mollastres}
	    <var> inmask <mask> parm <name> | parmindex <#> | <#> ... |
	  <var>=<start>;[<var><end OP><end>;]<var><increment OP>[<value>] ... }
	END KEYWORD: 'done'
  Create a 'for' loop around specified mask(s) and/or integer value(s).
  Any number and combination of masks and integers can be specified.
  Variables created in the for loop can be referenced by prefacing
  the name with a '$' character.
  Available 'end OP'       : '<' '>'
  Available 'increment OP' : '++', '--', '+=', '-='
```
Any number of loop types can be combined into a single `for` statement. For example:
```
	for atoms A0 inmask :1-27&!@H= i=1;i++
	  distance d$i :TCS $A0 out $i.dat
	done
```
Would loop over atoms in mask `:1-27&!@H=` while setting variable `i` to 1 and incrementing by 1 each iteration of the loop. If multiple end conditions exist within a single `for` loop the loop will execute until the first end condition is reached. For loops can also be nested.

Variables created by `for` loops are called "script" variables and are prefaced by `$`. This was chosen because it doesn't clash with existing cpptraj syntax and is familiar to almost all programmers. It also won't conflict with existing shell scripts since the `$` would need to be protected in order to make it to cpptraj (rendering it a non-variable to the shell script).

NOTE: the `for` loop could potentially render the commands `multicurve` and `multihist` obsolete. The `multidihedral` and `multivector` commands have some additional functionality which justifies keeping them around.

This also adds two commands that pertain to "script" variables.
```
> help set
  [help set]
	<variable>=<value>
  Set script variable <variable> to value <value>.
> help show
  [help show]
  Show all current script variables and their values.
```
This means that "script" variables can be used as placeholders in cpptraj scripts, e.g.:
```
set PREFIX = distance
for residues R0 inmask :1-4@O residues R1 inmask :9-12@H i=1;i++
  distance $R0 $R1 out $PREFIX.$i.dat
done
```
Note that "script" variables in `for` loops themselves are currently only allowed in the outermost loop; script variables in nested loops are not currently supported.


drroe:
The jenkins failure is a cuda driver problemI think, not code.

Updated syntax for `set`:
```
	{ <variable> <OP> <value> |
	  <variable> <OP> {atoms|residues|molecules} inmask <mask>
	    [parm <name> | parmindex <#> | <#>]
	  <variable> <OP> trajinframes }
  Set (<OP> = '=') or append (<OP> = '+=') a script variable.
  - Set script variable <variable> to value <value>.
  - Set script variable to the number of atoms/residues/molecules in
     the given atom mask.
  - Set script variable to the current number of frames that will
     be read from all previous 'trajin' statements.
```
Scalar data sets can now be used as script variables as well, although only the first value from a data set will be used. This can be expanded later if necessary.

PR #568, opened 2017-11-14
Fix MKL + GNU compile.

drroe:
Fixes for serial and OpenMP compiles. Should also fix PGI compile.


hainm:
Just FYI in case you need MKL on travis:

https://anaconda.org/intel/mkl

drroe:
@hainm Thanks Hai

PR #571, opened 2017-12-18
Add CMake builds to continuous integration

multiplemonomials:
I was kind of worried that the standalone CMake build system for cpptraj was not getting tested regularly.  So, I added new Travis and Appveyor builds that use CMake.  It took a lot of commits, as always, but I finally got it working.  Both Appveyor and Travis now have `USE_CMAKE` environment variables that control whether the build script uses CMake or not.

Currently, the CMake system only links with `libsander` when building inside Amber.  So, the CMake tests don't build with the Sander API.  I'll try to fix that once I finally get around to making Amber use CMake's built-in pkg-config functionality.

For Appveyor, I had to move all the logic into batch files called by `appveyor.yml` since it now uses CMD if statements.  The CMake build generates a zip file artifact on Appveyor just like the configure build, except that CMake is able to bundle all the correct DLLs so that cpptraj can be run on platforms without MinGW or MSYS installed.  

If we can figure out how to get MKL on Appveyor (maybe through Anaconda?), I could set up a Visual Studio build on there as well.  Unlike the MinGW build, a VS build can be used with pytraj .  Of course VS lacks a Fortran compiler, but I set up the CMake system so that if you provide a prebuilt Visual Studio-built fftw (which I can build with my system), it doesn't ever request a Fortran compiler or compile pubfft, and you can build with Visual C++ alone.  Would that be a useful thing to have?

I also managed to get an OS X CMake build going.  Were you having issues with Travis failing with Mac builds?  I noticed the `shell_session_update()` workaround, and it took me forever to get that bug fixed.  I had to switch Travis to use the `xcode9.2` container image, and remove `set -e` from the build script.  Now though, the OS X cmake and non-cmake builds run properly.  I can disable them if you need though, since i noticed you said something about `libgfortran` in the Git log.

Finally, I fixed two more missing `<algorithm>` includes, and updated the CMake submodule to the latest and greatest version.

Happy holidays!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ matrix:
 #   - { os: osx, env: COMPILER=clang USE_CMAKE=0 BUILD_FLAGS="-macAccelerate --with-fftw3=/usr/local --with-netcdf=/usr/local -noarpack" BUILD_TYPE=install TEST_TYPE=test.showerrors }
 #    - { os: linux, env: COMPILER=gnu USE_CMAKE=0 BUILD_FLAGS="-openmp -shared" OPT=openmp OMP_NUM_THREADS=1 BUILD_TYPE=libcpptraj TEST_TYPE=test.libcpptraj }
 #    - { os: linux, env: COMPILER=gnu USE_CMAKE=1 BUILD_FLAGS="-DOPENMP=TRUE" OMP_NUM_THREADS=4 TEST_TYPE=test.showerrors}
-    - { os: osx, osx_image: xcode6.4, env: COMPILER=clang USE_CMAKE=1 BUILD_FLAGS="-DOPENMP=TRUE" OMP_NUM_THREADS=4 TEST_TYPE=test.showerrors}
+    - { os: osx, osx_image: xcode9.2, env: COMPILER=clang USE_CMAKE=1 BUILD_FLAGS="-DOPENMP=TRUE" OMP_NUM_THREADS=4 TEST_TYPE=test.showerrors}
 
 #shell_session_update() { echo "Overriding shell_session_update"; };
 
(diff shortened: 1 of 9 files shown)


multiplemonomials:
Looks like Jenkins is dying due to the same error I emailed Jason about:
```
make[2]: Entering directory '/iscratch/jenkins/jenkins/jenkins/home/workspace/Open Source Chemistry/cpptraj/COMPILER_FLAGS/-cuda/label/linux/src/cuda_kernels'
/opt/cuda/bin/nvcc  -c -arch=sm_30  -o core_kernels.o core_kernels.cu
In file included from /opt/cuda/bin/..//include/cuda_runtime.h:78:0,
                 from <command-line>:0:
/opt/cuda/bin/..//include/host_config.h:119:2: error: #error -- unsupported GNU version! gcc versions later than 5 are not supported!
 #error -- unsupported GNU version! gcc versions later than 5 are not supported!
  ^~~~~
```

drroe:
Wow - this is a ton of work. Great stuff!

> it doesn't ever request a Fortran compiler or compile pubfft, and you can build with Visual C++ alone. Would that be a useful thing to have?

I suppose more tests are better, but I don't know how many people are actually building cpptraj on windows. I could go either way.

> Were you having issues with Travis failing with Mac builds?

I think ultimately the reason we moved OSX builds to @swails Jenkins server is that it just took forever for OSX builds to start on Travis. I don't recall other issues - maybe @swails or @hainm do.

> Finally, I fixed two more missing <algorithm> includes, and updated the CMake submodule to the latest and greatest version.

Thanks! Happy holidays to you as well. Try to take some time off! :-)

multiplemonomials:
Oh by the way, I was looking in to ways to speed up the build of cpptraj, so I did an experiment to compile it using a CMake script that utilized a precompiled header to try and speed up compilation.  Sadly, with the time it took to compile the PCH, the compile times were identical almost to the second.  I guess it's because cpptraj has lots of small source headers instead of a few big ones.

PR #574, opened 2018-01-12
Fix Appveyor build failures with CMake

multiplemonomials:
I noticed that the Appveyor CMake build has been acting like a random number generator casted to boolean lately, and I finally figured out why.  It turns out that Appveyor is sometimes appending a random suffix to the build folder.  The Appveyor CMake build script needs this to have that path defined in a variable with forward slashes:
```bat
set SRCDIR=C:/projects/cpptraj
```

So, when cloned into `C:\projects\cpptraj-aof9y`, it was installing to the wrong place, and dying.

I just put a block into appveyor.yml that should force it to go to the right directory.

I also updated the CMake submodule, and fixed the scripts to be compatible with it.


drroe:
Cool, thanks.

PR #578, opened 2018-01-19
Add limited ability to manipulate parameters

drroe:
This PR gives cpptraj limited ability to read in parameters (specifically CHARMM parameters) and use them to modify those in an existing topology (`updateparameters`). This is arguably outside the scope of what cpptraj should be doing, however the `chamber` functionality in parmed currently appears to skip some CHARMM parameters (see [here](https://github.com/ParmEd/ParmEd/issues/939)). Since I'm not super familiar with parmed code and this needed to be done for a colleague ASAP, I implemented this in cpptraj (also reading the 80 MB topology takes parmed a while to do). Eventually I'd like to fix the parmed code; this is just a stopgap measure, which is why the command is hidden (also why no test case yet).


swails:
Sorry!! :(.  On my list of things to do.  Hopefully this weekend.

drroe:
> Sorry!! :(. On my list of things to do. Hopefully this weekend.

Not a problem :-). Just wanted to make it clear why cpptraj is venturing outside its scope in this PR.

swails:
For completeness, the issue with ParmEd was the fact that the PSF file is actually missing the Angle (and I presume dihedral) terms that ParmEd inserts a 0 into.

ParmEd adds these "extra" terms with 0-ed force constants as a hacky way (also the *only* way) to get the exclusion list right in Amber (which uses the bond, angle, and torsion lists to construct the atom pairs that are <=3 bonds separated) when converting from a file format that uses the bond graph to identify exclusion pairs and therefore doesn't always necessarily include explicit angle or dihedral terms for all pairs of atoms that form a valence angle or torsion angle with each other.  I have no idea how CHARMM behaves in this respect (and I suspect there may be some issue with the PSF generation and these terms should not have actually been dropped).

Getting that right is one of the many "gotchas" that trips up attempts to convert file formats.

drroe:
The appveyor failure is the same nonsense "hang for no reason" issue that hasn't happened for a while. Apparently it's back. Not sure why it happens for the CMAKE=0 build and not CMAKE=1 though.

PR #581, opened 2018-02-21
Add particle mesh Ewald calc via PME library from A. Simmonett

drroe:
The main point of this PR was to add a PME calculation to cpptraj, but it turned into a few more things. Addition of PME was accomplished via using a PME library developed by Andrew Simmonnett (@andysim) for the reciprocal part of the energy. Since the PME library required C++11 support (also FFTW3), the configure script needed to check for this. In adding the C++11 check I decided to rewrite the entire configure script to make it a bit more efficient, particularly in how it checks for and links in external libraries. It should also be smarter about determining whether C++/Fortran linking is required. I also bundled in ARPACK since it's a key library for CPPTRAJ but can be hard to obtain and compile compared to other libraries. And since I was adding a new energy term anyway, I decided to add a kinetic energy calculation as well and make some improvements to and clean up the existing `energy` action.

Summary:

1. Add PME calculation support to `energy`:
```
etype pme [cut <cutoff>] [dsumtol <dtol>] [order <order>]
          [ewcoeff <coeff>] [skinnb <skinnb>]
          [nfft <nfft1>,<nfft2>,<nfft3>] [erfcdx <dx>]
```
The `cut`, `dsumtol`, `ewcoeff`, and `skinnb` keywords mean the same as they do for `etype ewald`.
- **nfft** specify number of grid points for FFT in each direction.
- **order** B spline order.
- **erfcdx** Error function table delta (added for `etype ewald` as well).

2. Added KE calculation to `energy`:
```
[kinetic [ketype {vel|vv}] [dt <dt>]]
```
- **ketype vel** Calculate KE from on-step velocities.
- **ketype vv** (Velocity Verlet) Calculate KE from on-step forces and plus-half step velocities.
- **dt** Time step when using `ketype vv`
If no `ketype` is specified default is to use `ketype vv` when force/velocity info present, otherwise just use velocity info.

3. The `energy` action is now smarter about when only certain terms are desired. Also, if only one term is to be calculated the redundant 'Total' term is not calculated.

4. Rewrite of configure. C++11 support now enabled by default if available. If `-amberlib` is specified, FFTW3 support will be enabled if it is available in AMBERHOME. The output from configure has been cleaned up and a better summary of chosen options is written at the end. New command line options:
- **-noc++11** Disable C++11 support.
- **-<lib>** Enable support for library (without specifying path).
- **--compile-verbose** Turn on verbose compile messages.

5. The `box` action now correctly sets the internal box type.

6. In calculation expressions, allow operator characters inside data set aspects (e.g. `ENE[VDW1-4]`).

7. ARPACK now bundled with CPPTRAJ.


drroe:
Going to squash this merge since the original libpme header was a beast.

PR #582, opened 2018-02-21
Fixes and improvements for NetCDF reservoir creation

drroe:
Addresses bug where reservoir creation would fail if incoming trajectory contained temperature information (really this stems from my original poor decision reuse the temp0 variable name). This also slightly improves reservoir creation by automatically adding velocity/force information if present and not explicitly disabled with `noforce`/`novelocity`. Also prevents writing of unnecessary time variable to the reservoir.

Under the hood, the reservoir creation functionality is now in its own class (RemdReservoirNC) instead of piggy-backing off of Traj_AmberNetcdf.


src/RemdReservoirNC.h
```diff
@@ -14,7 +14,7 @@ class RemdReservoirNC : private NetcdfFile {
     int WriteReservoir(unsigned int, Frame const&, double, int);
     /// Close the reservoir
 #   ifdef BINTRAJ
-    void CloseReservoir(); { NC_close(); }
+    void CloseReservoir() { NC_close(); }
```
swails: semicolon

drroe: lol

PR #586, opened 2018-03-02
Fix up handling of implicit constant pH REMD data and ensemble file handling

drroe:
Because information for only a single residue is written for most steps in implicit solvent constant pH REMD simulations, the constant pH data was not being properly read. This PR fixes that. It also improves the processing of constant pH data in parallel - data from multiple ensemble members can be written to a single file via the new `ensextension` command. If `ensextension` is off, cpptraj will try to write all ensemble data to a single file (like the serial ensemble version default behavior). If `ensextension` is on, cpptraj will write ensemble data to separate files. 

TODO: `ensextension off` in parallel currently only works for the `cphstats` analysis.


swails:
wow

PR #588, opened 2018-03-03
FIX pytraj cloning error ("alert protocol version while accessing")

hainm:
The error happens because the test is using very old centos 5 docker image.
The fix is easy, just clone pytraj outside the image.

https://githubengineering.com/crypto-removal-notice/

@drroe Please squash this PR once you merge (from github button) to avoid my meaningless git comments.


drroe:
Thanks Hai

Issue #585, opened 2018-02-26
macos issue - ld: library not found for -lgfortran


hainm:
Hi,

pytraj has been using 
```
bash configure -shared -macAccelerate -noarpack   -nosanderlib  CXX="clang++ -stdlib=libstdc++"
```

but getting the above error recently.

Below command is just fine.
```
bash configure -shared -macAccelerate clang
```

drroe:
Does `./configure -shared -macAccelerate -noarpack -nosanderlib clang` work? You shouldn't need to sneak flags in via the CXX variable (at least not anymore).

The error message occurs because you are using the built-in FFT library (pubfft) which is written in Fortran, hence the need for the C++/Fortran link (`-lgfortran`), but Fortran libraries do not appear to be available on your system. Using `-noarpack` and specifying an external FFTW3 library will get rid of the Fortran dependency.

hainm:
> Does ./configure -shared -macAccelerate -noarpack -nosanderlib clang work? 

yeap.

>  but Fortran libraries do not appear to be available on your system. 

I think I have fotran libraries (If not, how specifying "clang" above works).

> You shouldn't need to sneak flags in via the CXX variable (at least not anymore).

How can I add "-stdlib=libstdc++"?

hainm:
(this is stand for standalone pytraj).

drroe:
@hainm Is this still an issue after #589?

hainm:
Hi, I will check this tonight.

hainm:
unfortunately this is still an issue. Below is exact command that works before.

```
bash configure -shared -macAccelerate -noarpack  -nosanderlib  \
     CXX="clang++ -stdlib=libstdc++"
```

drroe:
I can't reproduce this issue, so I think we will need to work a little closer to resolve this. You stated that
```
./configure -shared -macAccelerate -noarpack -nosanderlib clang
```
works, so can't you just use that?

hainm:
Do you mean cpptraj won't respect CXX and CC envs anymore? I need you to confirm those things before fixing things in pytraj.

The compilation is a bit complicated since pytraj have both standalone and ambertools versions, I don't want to mess it up before releasing AT.

drroe:
I finally figured it out. Sorry it took so long.

> Do you mean cpptraj won't respect CXX and CC envs anymore?

[...]

PR #591, opened 2018-03-13
Fix determination of Fortran-related linker flags

drroe:
PubFFT does not require Fortran linking flags.


drroe:
Should hopefully address #585.

PR #606, opened 2018-05-11
Add MSVC build in Appveyor

multiplemonomials:
Hey Dr. Roe!  I added a new AppVeyor build variant that uses MSVC.  First of all, it will regularly check compiler compatibility, and verify that there aren't any missing includes or other issues that prevent building with MSVC.  Also, it is the first step towards getting proper pytraj builds on Windows.  

As you probably know, MinGW and MSVC have incompatible C++ ABIs, so only code built with MSVC 2015 is able to be linked with any Python 3.6 extension modules.  The regular MinGW64 builds of cpptraj don't work.  The artifact produced on AppVeyor from this job, though, is a proper MSVC build suitable for linking to Python, so it's possible to build pytraj for Windows using it.  Instead of having to go through the whole process of getting cpptraj and its dependencies built, all one has to do it install MSVC2015, set `CPPTRAJ_LIBDIR` and `CPPTRAJ_HEADERDIR` to point into the AppVeyor artifact, and run pytraj's `setup.py`.  Then, a working Windows pytraj builds.

Normally of course cpptraj depends on Fortran and can't be built with MSVC alone.  However, I set up the CMake build system so that if you provide an external FFTW library, it doesn't ever request a Fortran compiler.  So, I built netcdf and fftw on my machine, converted the official OpenBLAS build to MSVC format, and packed all three into a dependencies file that gets downloaded from my Box account for each build.  So, the build system uses the provided FFTW, and doesn't ever ask for a Fortran compiler.

Making this work properly required a few code changes.  Because it's MSVC, all public code in the DLL needs to be explicily exported.  CMake automatically finds and exports functions and classes, but it is unable to automatically mark static class variables as exported.  So, all static class variables that are used by pytraj need to get `__declspec(dllexport)` markers.  I added a new header for this, `SymbolExporting.h`, and stuck `CPPTRAJ_EXPORT` delclarations on the 7 variables that need to be exported.  SymbolExporting.h does nothing when building with the configure script, but when triggered by macros defined by CMake, it configures the export and import declarations properly. When cpptraj is installed, CMake hardcodes the correct settings into the header, so programs linking to cpptraj don't have to use any special compile definitions or anything.

Unfortunately, I have not been able to get the cpptraj tests to run automatically in Appveyor; I keep running into some strange DLL-not found error that doesn't actually name the DLL that couldn't be found.  However, I've run the tests on my own machine with the AppVeyor artifact and they work fine, there appear to be only a few small rounding errors.  The test log is here: [test.log](https://github.com/Amber-MD/cpptraj/files/1994321/test.log).

One of the tests (RemLog) was initially crashing, but I figured it out: there was an unescaped percent sign in a printf in `Analysis_RemLog.cpp`.  Removing that fixed things, and the build seems to work great.

Let me know what you think!
Jamie


drroe:
Great work as usual Jamie. Everything looks pretty good to me (although I don't work/develop on windows so I can't realistically check anyway, heh).

PR #607, opened 2018-05-17
Add ability to save more than 1 representative structure per cluster

drroe:
Add the `savenreps <#>` keyword to the `cluster` analysis command. If this keyword is specified, the top `<#>` structures will be saved instead of only the top structure. This will slightly change the output format of the summary and info files when used. Each line of the summary file will now have `<#>` times 2 extra columns
```
Rep RepScore
```
Where `Rep` is the representative frame and `RepScore` is the score based on the current method for choosing representative structures. The `#Representative frames:` line of the info file will have format:
```
#Representative frames: { C0rep1 C0score1 ... C0rep<#> C0score<#> } { C1rep1 C1score1 ... C1rep<#> C1score<#> } ... { CNrep1 CNscore1 ... CNrep<#> CNscore<#> }
```
Where C0 is the top cluster and CN is the final cluster.


hdgsgsg:
I would like to use the 'savenreps' feature to get 30 representative structures with the 'cluster' command, but cpptraj does not seem to understand it. That is, it gives the following error message: [cluster] Not all arguments handled: [ savenreps 30 ]. Also, the 'help savenreps' command says 'No help found for savenreps '. I updated and recompiled my Ambertools18, but I could not get the 'savenreps' feature work so far. Could someone identify the problem? Thank you very much!

drroe:
The savenreps functionality hasn’t been ported to AmberTools yet (it will be in AT19). You’ll have to use the GitHub version for now.

hdgsgsg:
> The savenreps functionality hasn’t been ported to AmberTools yet (it will be in AT19). You’ll have to use the GitHub version for now.

I see, thank you!

PR #636, opened 2018-09-07
Add 'splitcoords' command and 'for VAR in <list>' for loop

drroe:
This PR adds a new command and new for loop type (and tests for both).

###  New splitcoords command
```
splitcoords <crd set> name <output set name>
```
Will split the trajectory `<crd set>` by molecule and output to a new COORDS set. I added this since I've been doing a lot of work with the REPLICA functionality in CHARMM lately and it's useful to be able to split the replicated system back into molecules. Very basic right now but gets the job done.

### New for loop type
```
for <var> in <list>
```
Here `<list>` is a comma-separated list of strings or file names; file names can include wild card characters (* or ?). So for example, you could do something like:
```
for TRAJ in mytraj.?.nc,mytraj.??.nc
  trajin $TRAJ
done
```
Here is a script that I've used combining both of these new features to split multiple CHARMM REPLICA PDBs up and calculate dihedral angles:
```
# First create all COORDS
parm diala_nrep*.psf

for IN in gmin.pdb,before.pdb OUT in Gmin,Before
  loadcrd $IN
  splitcoords $IN name $OUT
  crdout $OUT $OUT.mol2
  set CLIST += $OUT,
done

for t=1;t<12;t++
  loadcrd init_path_$t.pdb
  splitcoords init_path_$t.pdb name Init$t
  crdout Init$t Init.$t.mol2
  set CLIST += Init$t,
done

# Calculate phi/psi for all
for VAR in $CLIST
  crdaction $VAR multidihedral d$VAR phi psi
  dataset makexy d$VAR[phi] d$VAR[psi] name phipsi$VAR
done

writedata phipsi.agr phipsiBefore phipsiInit* phipsiGmin
```


hainm:
May be adding `new command` label rather `enhancement`. This would make it easier for me to search (vs going all the issues) to add support to pytraj.

PR #641, opened 2018-10-02
Add option to cache integer data sets to disk

drroe:
This PR is aimed at reducing data set memory usage in certain cases. The first data set to be used this way for testing is `DataSet_integer`, for use with the `hbond`/`lifetime` commands (addresses a user issue on Amber mailing list). Adds new command:
```
usediskcache {on|off}
```
When `usediskcache on` is specified, whenever an integer data set is allocated by the master DataSetList, it is cached to disk via a NetCDF file instead. This uses less memory at the expense of speed - seems the code is about 1-2x orders of magnitude slower when caching to disk. This will be a work in progress while I try to improve the speed somewhat.

A test is added for cached data sets.

This PR also adds a new keyword aimed at reducing memory usage in another way: for 1D standard data file reads you can use `onlycols` to only read certain columns from the file:
```
readdata <file> onlycols <range>
```
For example:
```
readdata myfile.dat onlycols 1,3-5 index 1
```
Will read columns 1, 3, 4, and 5, using column 1 as the index column. Note that if `index` is used in conjuction with `onlycols`, the index column must be one of the columns specified.


swails:
[ci-skip] only works for Jenkins on the amber repository, just FYI (and it has a - after the ci)

drroe:
> [ci-skip] only works for Jenkins on the amber repository, just FYI (and it has a - after the ci)

I think it works for Travis as well ([see here](https://reflectoring.io/skip-ci-build/)). At least it seems to be working in this case.

swails:
I stand corrected. :) Thanks

hainm:
I think this PR breaks pytraj build ( error in dataset integer).

https://travis-ci.org/Amber-MD/pytraj/jobs/505631275

Issue #372, opened 2016-07-02
Cpptraj header title - inconsistent 


hainm:
`generated` vs `Generated`

``` bash
 $ grep "Cpptraj generated" *
Action_CreateReservoir.cpp:    title.assign("Cpptraj generated structure reservoir");
Binary file Action_CreateReservoir.o matches
Traj_CharmmDcd.cpp:      SetTitle("Cpptraj generated dcd file.");
Binary file Traj_CharmmDcd.o matches
Traj_Mol2File.cpp:    SetTitle("Cpptraj generated mol2 file.");
Binary file Traj_Mol2File.o matches
Traj_SQM.cpp:    outTitle.assign("Cpptraj generated SQM input");
Binary file Traj_SQM.o matches
Traj_Tinker.cpp:    SetTitle("Cpptraj generated mol2 file.");
```

``` bash
$ grep "Cpptraj Gen" *
Traj_AmberCoord.cpp:      title.assign("Cpptraj Generated trajectory");
Binary file Traj_AmberCoord.o matches
Traj_AmberNetcdf.cpp:      SetTitle("Cpptraj Generated trajectory");
Binary file Traj_AmberNetcdf.o matches
Traj_AmberRestart.cpp:    outTitle.assign("Cpptraj Generated Restart");
Traj_AmberRestartNC.cpp:    SetTitle("Cpptraj Generated Restart");
Binary file Traj_AmberRestartNC.o matches
Binary file Traj_AmberRestart.o matches
Traj_NcEnsemble.cpp:      SetTitle("Cpptraj Generated trajectory");
```

hainm:
I raise this issue because I use those header to make sure `pytraj` writes to correct format.

drroe:
Can you just do a case-insensitive check? If it's a pain I can change the titles to be something more consistent, but this may avoid any similar issues in the future.

hainm:
> Can you just do a case-insensitive check

Yes.

>  If it's a pain I can change the titles to be something more consistent,

No, it's not a pain. Just a minor point for consistency.

drroe:
Addressed by #648

PR #648, opened 2018-11-30
Add simple XYZ trajectory format. Better version info for builds.

drroe:
This PR has several parts:

1. Adds simple XYZ trajectory format read/write. Trajectories can be very simply
```
ATOM X Y Z
```
or
```
X Y Z
```
optionally with a single header denoted by `#` or a header before each trajectory frame. The reader currently makes no assumptions about each frame size so the number of frames is considered unknown. May change this in the future. Also, the default output trajectory title has now been made consistent across formats (addresses #372). Adds a test for this format.
```
Options for XYZ: Keywords: xyz, Extensions: '.xyz'
	ftype {atomxyz|xyz}              : Choose either 'ATOM X Y Z' (default) or 'X Y Z' output format.
	titletype {none|single|perframe} : No title, one title (default), or title before every frame.
	width <#>                        : Output format width.
	prec <#>                         : Output format precision.
```

2. Better version information. Now during configure cpptraj will try to determine if the build is inside AmberTools or not; this will be noted when cpptraj is launched, e.g.
```
CPPTRAJ: Trajectory Analysis. V18.01 (GitHub)
```
or
```
CPPTRAJ: Trajectory Analysis. V18.01 (AmberTools)
```
In addition, if building from the GIT repository the current GIT hash will be noted by `--internal-version`
```
CPPTRAJ: Internal version # V4.10.1 GIT hash 425e77b080e02d848de7558a7716cee67cd70462
```

3. Enable the `as <format>` keyword for `trajin`.

4. Tightens up the automatic ID of Tinker/Amber ASCII trajectories.


hainm:
nice. I've waiting for this.

Issue #645, opened 2018-11-05
Allow the 'center1' and 'center2' keywords to be active at the same time in 'radial' command


drroe:
This will allow solvent-solvent RDFs where the solvent is a bit more complex than water. May require slight reorganization of the code.

drroe:
Addressed by #656.

PR #656, opened 2018-12-21
Enable RDF calculation by residue/molecule

drroe:
Adds 4 new options to the `radial` command:

- **byres1** : Use center of mass of residues selected by mask1
- **byres2** : Use center of mass of residues selected by mask2
- **bymol1** : Use center of mass of molecules selected by mask1
- **bymol2** : Use center of mass of molecules selected by mask2

In each case, the atoms used are only the atoms actually selected by the original mask. If only on byresX or bymolX option is specified, the other mask defaults to by atom. So for example:
```
radial out WatO-Trp.agr 0.5 10.0 :WAT@O :TRP byres2
```
Will calculate the RDF of all water oxygens to the center of mass of each TRP residue. The new modes are OpenMP-enabled like the rest of `radial`. However they are not currently compatible with `centerX` or `nointramol`.

Adds two new tests. This effectively addresses #645.


saikat0003:
I am very new to Github. How to get the code of radial?

drroe:
Either clone or download the GIT repository (both available via the green button on the [main page](https://github.com/Amber-MD/cpptraj). Then compile the code.

PR #657, opened 2019-01-08
Add ability to scale velocities to 'setvelocity'

drroe:
New options for `setvelocity`
```
scale [factor <fac>] [sx <xfac>] [sy <yfac>] [sz <zfac>] 
```
- **scale** - Indicate existing velocities will be scaled.
- **factor** - factor to scale velocities by.
- **sx** - factor to scale X component of velocity by.
- **sy** - factor to scale Y component of velocity by.
- **sz** - factor to scale Z component of velocity by.


swails:
```
setvelocity scale sx 0 sy 0 sz 10000
```

rocket

PR #664, opened 2019-02-01
Add first incarnation of parallel analysis, Hausdorff distance calculation

drroe:
This PR adds support for performing Analyses in parallel with cpptraj. Adds this command:
```
parallelanalysis [sync]
```
This command requests that all currently set up analyses be split up and run over currently available MPI processes. If `sync` is specified all data sets are synced back to the master thread for e.g. subsequent analysis.

The `parallelanalysis` command was added largely to support rapid calculation of multiple `2drms` analyses for use in creating a larger Hausdorff distance matrix, which introduces the new Hausdorff analysis command:
```
[help hausdorff]
	<set arg0> [<set arg1> ...]
	[outtype {basic|trimatrix nrows <#>|fullmatrix nrows <#> [ncols <#>]}]
	[name <output set name>] [out <file>] [outab <file>] [outba <file>]
```
Currently this command expects one or more 2D matrices which contain the distances between two sets; the Hausdorff distance is determined from this. The Hausdorff distance `H` is determined from:
```
H = max{dH(A,B), dH(B,A)]
```
Where `dH(A,B)` is the directed Hausdorff distance between sets A and B, etc. Colloquially speaking, the directed Hausdorff distance between A and B is determined as follows:
1. What is the closest approach (distance) of each point in A to any point in B?
2. Choose the largest distance from among those distances.

If desired, the output can be formed into a matrix, which can be useful e.g. when doing multiple 2D rms calculations on different regions of a trajectory.

This PR also adds a new keyword to gnuplot output, `title`, for changing the plot title.

Adds a few test cases for `hausdorff` and one for `parallelanalysis` + `hausdorff`.


hainm:
Should this issue labeled as `new analysis` too? pytraj missed it. :D

PR #666, opened 2019-02-14
Add support for Lennard-Jones PME to 'energy' command.

drroe:
This PR adds support for calculating LJ energy via PME - as with electrostatic PME the reciprocal part of the energy is handled by the HelPME library from Andy Simmonett (@andysim). Adds the following keywords to `energy`:
**ljpme**: Activates Lennard-Jones PME with default Ewald coefficient (0.4) - this is separate from the coefficient used for electrostatics.
**ewcoefflj <coeff>**: Activates Lennard-Jones PME with given Ewald coefficient.
**ljswidth <width>**: If specified, Lennard-Jones (PME or non-PME) will be determined using a switching function with given width.

The energy is properly corrected for Lorentz-Berthelot combining rules. Tested against the LJ PME energies output by CHARMM (implemented by Andy). The code could use some cleaning and optimization but it works.

This also implements a new command, `catcrd`, which can be used to concatenate COORDS data sets:
```
catcrd <set name0> <set name1> [<set nameX> ...] name <name>
```
This command proved very useful when testing the LJ PME implementation.

Also fixes bug where box was not being updated for COORDS data sets after using `box` command.

Adds tests for LJ PME and `catcrd`.


swails:
The PR of the beast!

drroe:
ironmaiden

Issue #675, opened 2019-03-05
Segmentation fault in SPAM when running `spampure` with solute present


drroe:
Reported by D. DasGupta on the Amber mailing list. Run that errors looks like this:
```
  [spam purewater name SPAM cut 12.0 out spampure.dat]
    SPAM:
	Calculating bulk value for pure solvent
	Printing solvent energies to spampure.dat
	Data set 'SPAM' index is water # * frame.
	Using a 12.00 Angstrom non-bonded cutoff with shifted EEL.
#Citation: Cui, G.; Swails, J.M.; Manas, E.S.; "SPAM: A Simple Approach
#          for Profiling Bound Water Molecules"
#          J. Chem. Theory Comput., 2013, 9 (12), pp 5539–5549.
  [go]
---------- RUN BEGIN -------------------------------------------------

PARAMETER FILES (1 total):
 0: ACN_IL1B_solvated.prmtop, 39605 atoms, 12325 res, box: Orthogonal, 12173 mol, 11955 solvent

INPUT TRAJECTORIES (1 total):
 0: 'first10.mdcrd' is an AMBER trajectory, Parm ACN_IL1B_solvated.prmtop (Orthogonal box) (reading 10 of 10)
  Coordinate processing will occur on 10 frames.

BEGIN TRAJECTORY PROCESSING:
.....................................................
ACTION SETUP FOR PARM 'ACN_IL1B_solvated.prmtop' (1 actions):
  0: [spam purewater name SPAM cut 12.0 out spampure.dat]
	Found 11955 solvent residues [WAT]
	Total Grid memory: 8.419 MB
TIME:	Pair List Setup: 0.0076 s
	Grid dimensions: 17 19 18 (5814 total).
----- first10.mdcrd (1-10, 1) -----
 0% free(): invalid size
./Test.sh: line 13: 10680 Aborted                 (core dumped) $CPPTRAJ -i run.in
```

drroe:
Fixed by #676.

PR #676, opened 2019-03-05
Fix SPAM pure water calc when solute present

drroe:
Used to segfault due to invalid indices. Address #675.


drroe:
Hm - merge button seems broken...

swails:
Looks like it works to me?

drroe:
Weird - worked now. Previously button wasn't functional after multiple refreshes and logins. Must have been a temporary hiccup in the GitHub web interface.

PR #687, opened 2019-03-13
Improve platform detection by configure

drroe:
This fixes incorrect detection of e.g. ppc64 and related platforms.


drroe:
Appveyor has been cued up forever. Given that this is a very minor change, I'm going to roll the dice and merge here (famous last words)...

PR #690, opened 2019-03-16
First steps towards pytraj testing as part of cpptraj merge gate

swails:
This adds a dockerfile that will create an image containing libcpptraj
properly set up to build pytraj. It also adds a step to the Jenkinsfile
to build and push the docker image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -126,10 +126,10 @@ pipeline {
                 echo "Building and pushing ambermd/libcpptraj:${env.DOCKER_IMAGE_TAG}"
                 script {
                     def image = docker.build("ambermd/libcpptraj:${env.DOCKER_IMAGE_TAG}",
-                                             "-f Dockerfile.libcpptraj ./devtools/ci/jenkins")
-//                  docker.withRegistry("", "amber-docker-credentials") {
-//                      image.push()
-//                  }
+                                             "-f ./devtools/ci/jenkins/Dockerfile.libcpptraj .")
+                    docker.withRegistry("", "amber-docker-credentials") {
+                        image.push()
+                    }
                 }
             }
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -90,6 +90,7 @@ pipeline {
                         dockerfile {
                             dir "devtools/ci/jenkins"
                             filename "Dockerfile.cuda"
+                            label "cuda"
                         }
                     }
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -132,7 +132,18 @@ pipeline {
                     }
                 }
             }
+        }
 
+        stage("Check pytraj with this version of libcpptraj") {
+            environment {
+                DOCKER_IMAGE_TAG = "${env.BRANCH_NAME == "master" ? "master" : env.ghprbActualCommit}"
+            }
+
+            steps {
+                build job: '/amber-github/pytraj',
+                           parameters: [string(name: 'LIBCPPTRAJ_IMAGE_TAG', value: "${DOCKER_IMAGE_TAG}"),
+                                        string(name: 'BRANCH_TO_BUILD', value: 'master')]
+            }
         }
     }
 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,6 +8,15 @@ pipeline {
 
     stages {
         stage("Build and test") {
+            when {
+                // Testing needs to be done as a merge gate. Since this is a
+                // somewhat pricey operation in the number of executors used,
+                // assume sufficient testing was already done for the master
+                // branch
+                not {
+                    branch "master"
+                }
+            }
             parallel {
                 stage("Linux GNU serial build") {
                     agent {
@@ -104,6 +113,26 @@ pipeline {
                     }
                 }
             }
+        }
+
+        stage("Build libcpptraj docker container") {
+            agent { label "linux && docker" }
+
+            environment {
+                DOCKER_IMAGE_TAG = "${env.BRANCH_NAME == "master" ? "master" : env.ghprbActualCommit}"
+            }
+
+            steps {
+                echo "Building and pushing ambermd/libcpptraj:${env.DOCKER_IMAGE_TAG}"
+                script {
+                    def image = docker.build("ambermd/libcpptraj:${env.DOCKER_IMAGE_TAG}"
+                                             "-f Dockerfile.libcpptraj ./devtools/ci/jenkins")
+//                  docker.withRegistry("", "amber-docker-credentials") {
+//                      image.push()
+//                  }
+                }
+            }
+
         }
     }
 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -125,7 +125,7 @@ pipeline {
             steps {
                 echo "Building and pushing ambermd/libcpptraj:${env.DOCKER_IMAGE_TAG}"
                 script {
-                    def image = docker.build("ambermd/libcpptraj:${env.DOCKER_IMAGE_TAG}"
+                    def image = docker.build("ambermd/libcpptraj:${env.DOCKER_IMAGE_TAG}",
                                              "-f Dockerfile.libcpptraj ./devtools/ci/jenkins")
 //                  docker.withRegistry("", "amber-docker-credentials") {
 //                      image.push()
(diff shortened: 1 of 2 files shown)


swails:
@drroe and @hainm -- I'm not exactly sure how we want to do this.  This PR adds the functionality that this pull request is built and pushed to dockerhub inside an image following the success of the various compiler tests.  Here are the options of what to do:

* Have the cpptraj merge gate fire off a pytraj test (on pytraj's master branch) and fail the build if pytraj fails.

The pros of this -- you are notified right away if pytraj breaks.  The failure of both the pytraj test and the cpptraj test can be configured to shoot out emails to the relevant interested parties.  The red X on the PR will alert you to a problem, and clicking on the Jenkins link will show you immediately that it's the pytraj test that failed.  If the cpptraj tests pass, you know it doesn't break pytraj.

The cons of this -- there will be no way to get the Jenkins tests for both pytraj and cpptraj working simultaneously.  The pytraj Jenkins test will always pull the docker image from cpptraj master, so when you submit a PR to fix pytraj for the API-breaking change in cpptraj, it will fail the Jenkins build for the PR where you fix it relative to the cpptraj PR in progress.  So you can get the pytraj tests to pass if you merge a failing cpptraj PR, or you can get the cpptraj tests to pass if you merge a failing pytraj build.  Note that pytraj will *also* fail its other tests since those also test against cpptraj master.  So really the only recourse in this case is to merge a failing cpptraj build and then let pytraj react.

* Have the cpptraj build fire off a pytraj test, but ignore its results when reporting success.  The pytraj test job that gets fired off will run and can be configured to alert (via email) anybody who wants to be notified that the pytraj test failed (we could probably change the email to say the test failed relative to a cpptraj pull request).

The pros of this - you still get notified when pytraj fails, but the Jenkins test only reflects whether cpptraj itself works with CUDA, Intel, MKL, PGI, MPI, OpenMP, etc.  You still get proactively notified if pytraj breaks, but there will be no indication through the Jenkins return status on cpptraj that pytraj was broken.

The cons of this - no priority is put on *not* breaking pytraj.  The expectation is that cpptraj can break things and pytraj needs to respond with fixes.  You can't tell immediately from the Jenkins status on cpptraj that it didn't break pytraj (but you will be able to tell if you click on the Jenkins status and drill into each step of the pipeline, but that's probably not typical behavior).

------

Which approach should we take here?

If the general policy is expected to be that cpptraj will *not* break pytraj, then we should opt for the first choice.  The build will fail and the cpptraj PR author (usually @drroe) will be able to drill down to see why pytraj broke and fix the cpptraj API (and the Jenkins test will not pass until pytraj passes).

If the general policy is that cpptraj will try not to break pytraj, but that occasional breaks are acceptable if it is a consequence of a design improvement, then I would still opt for the first choice.  We would merge the occasional failed Jenkins build, but we'd be proactively notified of pytraj breaks and be able to decide whether the breakage is worthwhile.

If the expectation is that these breaks will not be reasonably common and that pytraj will just need to keep up, then I'd recommend going with the second option.  @hainm will still be notified of breaks, but the cpptraj build won't be held up by pytraj breaks.

However, if the expectation is that breaks will be rare, and @drroe wants to hear about them to see how easy it would be to fix before merging a PR into master, then I'd recommend the first option.

The more I think about it, the more I think the first option is the way to go, but I'm open to counterarguments (it's really up to @drroe who does the bulk of the development work here, and cpptraj develops and is used at a much higher clip than pytraj, ParmEd, or anything else on GitHub).

drroe:
Thanks for this @swails. I’m feeling a bit under the weather today so I don’t know if I’m grokking the options. I think the second one seems a little like the way we had been doing things, so maybe option 1 seems right, but I’ll wait for @hainm to weigh in.

hainm:
yeah, let's follow option 1. I prefer to have Dan be aware of the issue of breaking pytraj build before merging cpptraj. e.g: Dan has pushed https://github.com/Amber-MD/cpptraj/pull/641 that broke pytraj build while we're very close to code frozen in AT.

devtools/ci/jenkins/Dockerfile.libcpptraj
```diff
@@ -0,0 +1,23 @@
+FROM continuumio/miniconda:latest
```
hainm: should we make a base image (saying: `ambermd/libcpptraj-base`) using line 1 to 14 first (that is built based on continuumio/miniconda:latest but do all `apt-get` and 
```
                sh "conda install -y --file conda-requirements.txt"
                sh "pip install -r pip-requirements.txt"
```

Then

```
FROM ambermd/libcpptraj-base
COPY . /cpptraj

 ENV CPPTRAJHOME "/cpptraj"

 RUN cd /cpptraj && \
    ./configure -openmp -shared gnu && \
    make -j4 libcpptraj
```

So we don't need lie 27-2 in this file: https://github.com/Amber-MD/pytraj/pull/1482/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3R27

The aim is that we only let pytraj build fail because of cpptraj itself.

hainm: (Or I just overthink)

Jenkinsfile
```diff
@@ -105,5 +115,36 @@ pipeline {
...
+
+        stage("Build libcpptraj docker container") {
+            agent { label "linux && docker" }
+
+            environment {
+                DOCKER_IMAGE_TAG = "${env.BRANCH_NAME == "master" ? "master" : env.ghprbActualCommit}"
+            }
+
+            steps {
+                echo "Building and pushing ambermd/libcpptraj:${env.DOCKER_IMAGE_TAG}"
```
hainm: How many images we will keep? (each image costs ~700M).

devtools/ci/jenkins/Dockerfile.libcpptraj
swails: Sounds like a bigger maintenance headache to me than anything else.  This image serves *one* purpose, which is providing a pre-packaged cpptraj install set up specifically to support testing pytraj against a cpptraj commit inside Jenkins.

That's why I'm comfortable doing a whole bunch of `chmod`s and leaving a suboptimal/dirty environment in place.  I don't expect this image to actually be *used* for anything else.  Nor do I think the `ambermd/libcpptraj` image will be of general interest.

If the base you propose making turns out to be generally helpful to more than just this image, I'd support making it.  But until then, I'd rather not have to maintain *both* `ambermd/libcpptraj-base` *and* this one.  But I don't really fancy the idea of maintaining 2 images for only this use-case.

Besides, the current `Dockerfile.libcpptraj` is only 22 lines.  Yours is 8.  Seems like a lot of work to save 14 lines of trivially understandable `Dockerfile`.

Jenkinsfile
swails: Yea, I don't have a great answer for this.  Ideally the only one I want to keep long-term is the `master` branch.  The PR-specific tags I'd like to get rid of, but I don't have a good way other than manually purging them periodically.  It's also unclear to me how long some of those PR-specific images will be useful, particularly if it breaks pytraj.

It isn't all that bad, though.  Each image involves ~700M of download, but the images themselves should only require a small amount of extra storage.

Docker images are composed of multiple layers that are put together to construct a full container.  Each line of the Dockerfile will generate a new layer (which is a function both of the layers it is built on top of as well as the command that creates that new layer).

Since the bulk of the image is the same for *every* `ambermd/libcpptraj` image -- specifically the part that comes at the top -- those layers will be the same for every `libcpptraj` image and thus stored only once.

The only layers that change are the ones that contain the compiled cpptraj source code.  So all these images won't actually take up that much space on Dockerhub.

swails:
I think this is ready to merge unless more issues come up

drroe:
@swails I'm ok with merging this. However I think I should do a cpptraj release for AT19 before I merge this since it is unlikely this PR will get merged into the AmberTools tree - unless you think I should push for that.

swails:
Sounds reasonable. The only benefit of this pr is that it alerts about pytraj breakages. That’s not really a risk here if we aren’t doing any real changes.

swails:
When you do merge it, I want to make sure that it pushes a libpptraj docker image with the master tag.

hainm:
@drroe  ... or you can make at19 branch for current AT and then merge this PR (if that's not complicated for you).

swails:
Tag would be better. 

You can always branch from the tag later if there’s a need. 

Let’s just have this PR be the first one merged after the release. Pytraj won’t break between now and then and we won’t mess up Dans release process.

drroe:
On second thought, I think I'm going to merge this and try to get it into the AT release since I'm currently working on a bugfix PR that I also want in the release. @swails do you see any downsides to merging?

swails:
Nope. Merge away. 

merge

PR #704, opened 2019-04-03
Quaternion Fix and add ions to the solute atoms group

acruzpr:
There is a bug in the calculation of quaternions for the orientational entropy calculations.   This was fixed by changing two lines of code.  The resulting errors in the entropy were small (< 1%) but this is fixed with the changes.  

Also, there was code that did not calculate ion-water interactions in the GIST energies.  This was likely put in to discount ion-water interactions from ions that were added to the simulation to neutralize the system however, the code also ignored single ions in the active site which should be considered part of the protein (e.g. 2FBX with two Magnesium atoms in the active site).


swails:
Can one of the admins verify this patch?

src/Action_GIST.cpp
```diff
@@ -393,13 +393,13 @@ Action::RetType Action_GIST::Setup(ActionSetup& setup) {
       isFirstSolvent = false;
     } else {
       // This is a non-solvent molecule. Save atom indices unless 1 atom (probably ion).
-      if (mol->NumAtoms() > 1) {
+      //if (mol->NumAtoms() > 1) {
```
drroe: Maybe instead of turning it off outright, we can add a keyword called `includeions` or something similar that will have the same effect, e.g.
```C++
if (mol->NumAtoms() > 1 || includeIons_) {
```
this will preserve backwards compatibility while making the routine more flexible.

drroe:
@acruzpr can you please verify that the updated numbers being produced by the GIST test are correct, and if so can you update the test save? Thanks.

src/Action_GIST.cpp
acruzpr: Thanks Dan, that looks great.  I (Tom here) would like includeIons to default to true as I think that most users will just run the defaults and not be aware that relevant ions that are effectively part of the protein will be ignored if the default is false.

drroe: >  I (Tom here) would like includeIons to default to true as I think that most users will just run the defaults and not be aware that relevant ions that are effectively part of the protein will be ignored if the default is false.

Sounds fine. When the change is made can you also increment the minor version number (second number) in `Version.h` since this will alter results from what was previously output from cpptraj GIST? Thanks!

acruzpr:
@drroe The only file in the GIST test that needed an update was the gist.out.save because it includes the results of the orientational entropy calculation. I updated it already.

Thanks for all your help.

drroe:
@swails The jenkins build didn't trigger - any idea why?

drroe:
run jenkins

swails:
> @swails The jenkins build didn't trigger - any idea why?

It's a security feature.  Jenkins won't run any pull request not authored by someone on my whitelist.  My whitelist is basically everyone in the `Amber-MD` organization.

This is intentional because Jenkins is basically the *ultimate* in the arbitrary code execution vulnerability :smiley:

I've done my best to sandbox the Jenkins execution environments (for instance, all slave nodes execute as a permission-limited user), but there's only so much I can do without crippling the ability to run tests.  For instance, the Jenkins user accounts need access to potentially vulnerable components of the underlying kernel -- like the video driver (otherwise CUDA tests would be blocked).

Anybody could submit a pull request, and since the build recipe is stored in code (via the `Jenkinsfile`), they can arbitrarily modify it to run whatever they want.  If tests executed automatically for everyone, a malicious developer could cause significant damage to my underlying machines.

The end result is a policy that whitelists people I know and trust and grants them the ability to trigger builds for PRs that they trust (as you did in this instance).

If someone becomes an active cpptraj contributor, you can always add them to the whitelist in Jenkins in the cpptraj merge gate job configuration.  I can show you how to do that if you want.

drroe:
@acruzpr I've implemented an `excludeions` option for GIST that restores the previous behavior - the new default is to include ions. I've also bumped the minor version number. Once the CI tests pass this should be good to merge.

drroe:
run jenkins

PR #719, opened 2019-05-09
Add rules for building static libraries.

drroe:
Adds 4 static libraries:

* **libcpptraj_core.a** : Contains core functionality; basic IO for STDOUT/STDERR, argument lists, topology, etc.
* **libcpptraj_file.a** : Basic File IO.
* **libcpptraj_traj.a** : Single trajectory file IO.
* **libcpptraj_parm.a** : Topology file IO.

This is currently so that `ambpdb` in AmberTools can avoid using the shared library, but may be useful for other programs down the road.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,10 @@ install: config.h
 libcpptraj: config.h
 	cd src && $(MAKE) libcpptraj
 
+# Create static libraries
+libstatic: config.h
+	cd src && $(MAKE) libstatic
+
 # Run Tests
 check: config.h
 	cd test && $(MAKE) test.complete summary
--- a/src/Makefile
+++ b/src/Makefile
@@ -40,8 +40,8 @@ $(CPPTRAJLIB)/libcpptraj_core.a: $(LIBCPPTRAJ_CORE_OBJECTS)
 $(CPPTRAJLIB)/libcpptraj_file.a: $(LIBCPPTRAJ_FILE_OBJECTS)
 	ar rv $(CPPTRAJLIB)/libcpptraj_file.a $(LIBCPPTRAJ_FILE_OBJECTS)
 
-$(CPPTRAJLIB)/libcpptraj_traj.a: $(LIBCPPTRAJ_TRAJ_OBJECTS)
-	ar rv $(CPPTRAJLIB)/libcpptraj_traj.a $(LIBCPPTRAJ_TRAJ_OBJECTS)
+$(CPPTRAJLIB)/libcpptraj_traj.a: $(LIBCPPTRAJ_TRAJ_OBJECTS) $(XDRFILE_OBJECTS)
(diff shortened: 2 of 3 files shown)


hainm:
can we have a `libcpptraj.a` too?

drroe:
> can we have a libcpptraj.a too?

What exactly do you need it for? Knowing this will allow me to tailor it so that only the essential stuff gets put into a library, and things that you might not care about (like the interactive mode) are not. The ideal thing would be if you provide me a file that needs to link libcpptraj.a so I can see what symbols are needed.

hainm:
ah, I just want to try with pytraj. But it's ok if there is not, pytraj works quite well now. :D

drroe:
> ah, I just want to try with pytraj. But it's ok if there is not, pytraj works quite well now. :D

I don't mind working on it - I just think that if we're going to do it let's make sure we get everything pytraj needs, but only what it needs. One thing you could do would be to try to use the existing static libraries I created and then tell me what symbols are missing. Off the bat you'll be missing all the Action and Analysis symbols, so I could work on that, but I'd be interested as to what else is missing.

Of course, we could also just say "forget it" and link everything into a mega-library like the existing libcpptraj.so :-).

drroe:
@swails Appears there might be a problem with the GPU on the Jenkins runner.

hainm:
> I don't mind working on it - I just think that if we're going to do it let's make sure we get everything pytraj needs, but only what it needs

> Of course, we could also just say "forget it" and link everything into a mega-library like the existing libcpptraj.so :-).

Thankks, but yeah, just "forget it". It already works well. 

>  One thing you could do would be to try to use the existing static libraries I created and then tell me what symbols are missing. Off the bat you'll be missing all the Action and Analysis symbols, so I could work on that, but I'd be interested as to what else is missing.

I will (but I am not sure about time frame though).

swails:
run jenkins

swails:
I updated my Gentoo system, which must have updated the nvidia drivers.  System needs a reboot.  But that node is the one hosting Jenkins, so I'm going to wait until all jobs finish on it.

In the meantime, I've brought my other GPU node back online (I already finished updating and rebooting that one) and marked the one that needs rebooting as unavailable so it won't get used again until it's ready.

dacase:
Propsed update looks fine to me.  Better than what I had thought of doing.

Is it the case the libcpptraj.so can be skipped if one is not building pytraj?  That is, does anything else use it (like cpptraj itself)?  I'm thinking the answer is "no", since I seem to be able to build a static version of cpptraj.

drroe:
> Is it the case the libcpptraj.so can be skipped if one is not building pytraj?

Yes - once ambpdb is modified to use the static libraries generated in this PR, pytraj is the only thing left that will need libcpptraj.so.

drroe:
@dacase I'll work on getting this incorporated into the Amber gitlab today.

PR #732, opened 2019-06-13
Fix potential overflow in string dataset

drroe:
When determining if printf should be used, check buffer size against actual format width, not just the data size.


drroe:
run jenkins

drroe:
Note sure what's up with Jenkins. I'll try doing another commit and see if it triggers.

drroe:
Jenkins doesnt appear to be working today. Merging anyway.

PR #739, opened 2019-07-11
Add ability to write CHARMM coordinate files.

drroe:
Previously, could only read.

```
  [help Formats trajout cor]
    Write options for Charmm COOR: Keywords: cor; Extensions: '.cor'
	keepext                : Keep filename extension; write '<name>.<num>.<ext>'
	ext                    : Use 'extended' format (default when > 99999 atoms).
	segid <segid>          : Use <segid> as segment ID for all atoms.
	segmask <mask> <segid> : Use <segid> as segment ID for atoms selected by <mask>.
```


Olllom:
This is super helpful, Dan. Thanks! 
A minor suggestion: writing out the periodic box vectors in the header comment could be a good idea. Only if it is not too much work of course.

hainm:
> A minor suggestion: writing out the periodic box vectors in the header comment could be a good idea.

@Olllom Just curious: why is that helpful?

Olllom:
In my opinion the box size should be "a part of" the coordinate set.
The only reason for me to convert files into the .cor format is to start
calculations in CHARMM.
So why not also specify the box if it is known at the time of writing the
file so that people don't have to guess when setting up the CRYSTAL.

Anyway, it's a very minor comment. And I don't mind if it is not
implemented.



Am Do., 5. Dez. 2019 um 16:20 Uhr schrieb Hai Nguyen <
notifications@github.com>:

drroe:
> writing out the periodic box vectors in the header comment could be a good idea. 

It's not a problem. Can you give an example of the format you would prefer?

Olllom:
Here's an example of what openmm writes into the pdb header:
```
REMARK   1 CREATED WITH OPENMM 7.3, 2019-02-28
CRYST1   30.584   30.584   30.584  90.00  90.00  90.00 P 1           1 
```
Maybe that would work.

Issue #735, opened 2019-06-17
Amplitude values for pucker are reported in radians


drroe:
Should probably be converted to degrees. Double check this!

drroe:
Addressed by #740.

PR #740, opened 2019-07-20
Miscellaneous Fixes

drroe:
1. Actually use the value passed in to the `DataSet_integer_mem::Assign()` function - doesn't currently cause bugs bug could down the road.
2. Convert calculated pucker amplitude values to degrees to be consistent with pucker and theta. Addresses #735.
3. Update documentation for `pucker` (clarifying all values are in degrees) and `distance` (clarifying how `reference` is used).
4. Fix parsing of PSF files with alpha characters in the residue ID field (which is legal); related to ParmEd/ParmEd#1058. Code will attempt to use alpha character found as residue insertion code.

### DIFF
--- a/doc/cpptraj.lyx
+++ b/doc/cpptraj.lyx
@@ -28999,8 +28999,8 @@ amplitude
 \series bold
 theta
 \series default
- keywords are given, amplitudes/thetas will be calculated in addition to
- pucker.
+ keywords are given, amplitudes/thetas (also in degrees) will be calculated
+ in addition to pucker.
  The results from 
 \series bold
 \shape italic
--- a/doc/cpptraj.lyx
+++ b/doc/cpptraj.lyx
@@ -20385,8 +20385,8 @@ refindex
 \begin_inset space ~
 \end_inset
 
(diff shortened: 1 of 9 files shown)


swails:
How does your charmm coordinate parser handle residue indices with letters? That was parmeds problem.

drroe:
> How does your charmm coordinate parser handle residue indices with letters? 

@swails So it's semi-hacky. Cpptraj has always used the the PSF residue ID field as "original" residue number, i.e. the residue number in the file, which does not necessarily correspond to the internal residue number (because e.g. some PDBs start at residues other than 1 etc). So what cpptraj does is read the residue ID string until it runs out of digits and considers that the original residue number, Then if there is an alpha character after that, it makes that the residue insertion code. Now I don't know if that's really what it's supposed to be, but it seemed potentially OK and was a simple way to preserve the information. If the residue ID string contains no digits cpptraj will complain and give an "original" residue number of 0, which may also not be great but I wasn't sure what the best option was there. It seems to work ok for the given test case, but may need further behavior changes down the road.

PR #741, opened 2019-07-25
Fix potential segfault in calcstate

drroe:
If transitions file name not specified, would segfault due to null pointer.

Also update the manual PDF.


drroe:
Jenkins issue - will try to re-run.

drroe:
run jenkins

drroe:
@swails seems the jenkins server is having issues:
```
 > git ls-remote -h git@gitlab.ambermd.org:/swails/jenkins-shared-libraries.git # timeout=10
ERROR: Checkout failed
hudson.plugins.git.GitException: Command "git ls-remote -h git@gitlab.ambermd.org:/swails/jenkins-shared-libraries.git" returned status code 128:
stdout: 
stderr: git@gitlab.ambermd.org: Permission denied (publickey).
fatal: Could not read from remote repository.
```
Since everything else passes and this is a critical fix I'm going to go ahead and merge.

swails:
That's what I get for experimenting.

Should be fixed now.  I was experimenting with shared pipeline libraries in Jenkins (cool stuff).  But I was done with experimenting and got rid of the shared library repo on gitlab I was using, which is why the checkout failed.  Since these shared libraries are basically auto-imports into every Jenkins pipeline script, all pipelines broke when I removed the repo.

I deleted the pipeline library import inside Jenkins.  Should be good now.

PR #751, opened 2019-09-10
Remove all references to '/dev/*' in test framework.

drroe:
Related to #750 and some of the changes in #747. File redirection in scripts should now be properly handled everywhere.

### DIFF
--- a/test/MasterTest.sh
+++ b/test/MasterTest.sh
@@ -267,10 +267,11 @@ GetResultsFiles() {
     done
   else
     if [ "$1" = 'single' ] ; then
-      RESULTSFILES=`ls $2 2> /dev/null`
+      RESULTSFILES=`ls $2 2> tmp.cpptrajtest.devnull`
     else
-      RESULTSFILES=`ls */$2 2> /dev/null`
+      RESULTSFILES=`ls */$2 2> tmp.cpptrajtest.devnull`
     fi
+    rm tmp.cpptrajtest.devnull
   fi
 }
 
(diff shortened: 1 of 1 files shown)


drroe:
I don't get PGI compilers. Last Jenkins run (91) passed, now this one (92) fails. It's very arbitrary and I'm really inclined to just ignore it.

swails:
If we're going to ignore the PGI fileaures (which I totally support), then we should remove that stage from the Jenkinsfile.

Keep in mind that Jenkins *also* is responsible for detecting pytraj breaks, but it only does that if all of the individual stages succeed.  So each time PGI fails and we ignore it, pytraj build/tests never get run with the proposed changes.

drroe:
> If we're going to ignore the PGI fileaures (which I totally support), then we should remove that stage from the Jenkinsfile.

Long term I agree. I'd like to keep it for now to see if the failures really are random. If over the next few PR I get sometimes passing, sometimes failing I'll consider it too much of a pain to debug and disable it. However, if it's failing consistently I'd like to address that.

swails:
Well at least we shouldn't block the pytraj tests from proceeding if the PGI builds fail.

sbrozell:
Keep me,  @sbrozell , in the loop.
For context see
https://gitlab.ambermd.org/amber/amber/merge_requests/319

sbrozell:
@drroe please include me in your repo; not sure what i need to do ... thanks

hainm:
@sbrozell you can click "Watch" in the top page.https://help.github.com/en/articles/watching-and-unwatching-repositories

PR #752, opened 2019-09-10
Do not make PGI builds fatal

swails:


drroe:
Appveyor failure is the usual timeout - looks good to me.

PR #768, opened 2019-10-18
Add support for reading Gromacs TNG trajectory files

drroe:
Will read coordinates, box info, velocities, and forces from Gromacs TNG files. Makes use of the Gromacs TNG library (https://github.com/gromacs/tng), which will be bundled with cpptraj. Adds a test and updates manual/README.


hainm:
OMG, for real?

drroe:
For some reason the tng subdirectory was still looked at by lgtm. Going to merge anyway and deal with that later.

PR #772, opened 2019-11-07
Update Jenkinsfile

swails:
* Do not use dockerfiles anymore -- pull images directly from dockerhub
that are created by the common-dockerfiles repo
* clone/stash/copy instead of clone each time
* cleanup after each stage


swails:
This removes the dependency on the GHPRB plugin which I think is to blame for the recent breakage in Jenkins.

PR #773, opened 2019-11-11
Build documentation and serve them (for the time being) from Jenkins

swails:


swails:
This is a general proof-of-concept.  But you can see the result here:

* Cpptraj Manual is built and available through Jenkins: https://jenkins.jasonswails.com/job/amber-github/job/cpptraj/job/PR-773/ -- see the two artifacts which are the PDFs themselves.
* doxygen docs are generated and published through Jenkins' HTML publisher (look for "Doxygen" in the side-bar of the link above).

This way people always have the latest documentation from the latest master branch (and you can look at the resulting PDF and doxygen docs for each branch/pull request).  It'll also gate on bad changes to the lyx file.

Let me know what you think.

drroe:
> Let me know what you think.

I think it's outstanding, thanks! I never liked tracking the PDFs through GIT since as you know GIT isn't so happy with binary files and lots of times when I updated the manual LyX source I would forget to regenerate the PDF. This is much more elegant.

Is this GtG? If so I'll go ahead and merge.

swails:
The links might not work in the readme file. I tried to predict what they'd be, but the only way I'll know for sure is to merge it and fix any mistakes. So go ahead and merge.

PR #776, opened 2019-11-13
add fabs and 8 more voxels for NNr

EricChen521:
1) Search 27 voxels for Translational and Six dimensional Entropy

2) Add fabs for quaternion distance calculation


drroe:
@EricChen521 thanks for this!

Your changes significantly alter the output for the GIST tests, which is why CI is failing. Please double check all calculations to make sure they are correct. When you are sure the changed output is something you expect, update the test output. (Pinging @tkurtzman for his input on this as well).

EricChen521:
Hi Daniel,

Thank you very much for testing the updated code! We fixed a few bugs, so the output will surely be different from that of the previous version. As you suggested, I have updated the output file for future testing purposes.

drroe:
@EricChen521 thanks! Looks good now - can you just bump the revision number (last number in Version.h) so that it's 4.23.3 - thanks! After that it will be good to commit

drroe:
Sorry for the delay on this - have been extremely busy. Will merge today.

PR #785, opened 2019-12-13
Fix 'temperature' action for removed global degrees of freedom

drroe:
Adds the `remove` keyword so that temperature calculation can account for removed translational and/or rotational degrees of freedom (e.g. an implicit solvent trajectory with CoM motion removed). Also adds an `update` keyword so the in-frame temperature can be updated (similar to the `time` action). Adds a test, updates docs.

Also adds a small section in the developers guide about what to do when modifying Topology/CoordinateInfo in Action::Setup().


drroe:
Pytraj build on Jenkins is taking a while today...

hainm:
omg, more than 5 hours. Not sure what happened.

drroe:
Ok....I'm considering just merging this, but I'd like to know what happened with pytraj on Jenkins. @swails any insights?

swails:
Looks like it just hung on casegroup...

PR #787, opened 2020-01-08
Interaction Energy calculation on GPU for GIST

jokr91:
- Added GPU capability for the calculation of the interaction energy in GIST.
- Also added a skipS flag, in order to skip the entropical calculations for GIST (calculate only the interaction energies).
- Added a few sentences in the documentation, concerning the GPU implementation. Added the skipS flag at the beginning of the "gist" command.
- In the help function, I added the skipS flag.
- Pushed the version from 4.24.0 to 4.25.0.
- For testing, added a "cuda" and "notcuda" keyword for the test runs.


src/cuda_kernels/ConstantsG.cuh
```diff
@@ -0,0 +1,69 @@
+/**
+ * This is simply taken from the cpptraj implementation by Daniel R. Roe and Mark J. Williamson.
```
drroe: @jokr91 Why not just refer to Constants.h in the `src` directory?

src/cuda_kernels/GistCudaSetup.cu
```diff
@@ -0,0 +1,239 @@
+#include "GistCudaSetup.cuh"
+#include "GistCudaCalc.cuh"
+#include <iostream>
```
drroe: I had to change this from `<iostream>` to `<cstdio>` in order to avoid:
```
GistCudaSetup.cu(195): error: identifier "printf" is undefined

1 error detected in the compilation of "/tmp/tmpxft_00002bc7_00000000-5_GistCudaSetup.cpp4.ii".
```
on my local machine (gcc 5.4, cuda 8)

drroe:
@jokr91 This is great work, thanks!

I had two minor comments (above). Also, could you add `gist` to the list of CUDA-enabled commands in section 2.7.3 of the manual? After that, looks good to merge!

src/cuda_kernels/ConstantsG.cuh
jokr91: Sorry, I removed this file.

src/cuda_kernels/GistCudaSetup.cu
jokr91: I changed the `<iostream>` to `<cstdio>` and moved the `#include <cstdio>` into the GistCudaSetup.cuh. Should now work, but could you please double check?

drroe:
Jenkins failure is a problem with the build framework. Since I've gotten the CUDA tests to pass locally I will merge. @jokr91 after I merge can you please re-test to make sure I didn't miss anything? Thanks.

drroe:
I'm not going to wait for the appveyor tests.

jokr91:
> 
> 
> Jenkins failure is a problem with the build framework. Since I've gotten the CUDA tests to pass locally I will merge. @jokr91 after I merge can you please re-test to make sure I didn't miss anything? Thanks.

I ran the CUDA tests on my local machine as well and they are passing.

PR #788, opened 2020-01-09
Fix GIST order calculation for water models with more than 3 points

drroe:
A simple but necessary fix. Adds tests for order calculations with tip4p and tip5p water models to catch any issues like this in the future. Should be merged before #787. Not incrementing the version number since #787 (which should be merged right after) already increments the minor version. #787 should be fully compatible with this PR.

@jokr91 I'll merge this before #787 just so the tests get in the tree, but I already checked that your new GPU code gives the right answers (which is awesome!). It's just the CPU code that had the issue.


drroe:
The Jenkins failure is for the CUDA build but is framework=related, not code:
```
Failed to run image 'ambermd/gpu-build:latest'. Error: docker: Error response from daemon: OCI runtime create failed:
```

Issue #744, opened 2019-08-02
RPN calculator mistakes variables as functions if variable starts with function name


drroe:
E.g., the variable `sumTotal` will be mistakenly identified as the `sum()` function.

drroe:
This was addressed way back in #789

PR #789, opened 2020-01-10
Major improvements to loops, misc other fixes

drroe:
This is a pretty big PR that provides major improvements to the utility of loops, as well as other changes under the hood:

- `for` loops have been fixed so that nested loops can make use of variables, including outer loop variables. For example, the following will now work:
```
set maxval = 4
set maxval1 = 3
set val = 0
for i=1;i<$maxval;i++ myval=0;myval+=10
  startj = $i + 1
  for j=$startj;j<=$maxval1;j++
    distance d$\$j @$i @$j out nested.agr
  done
done
```
This will result in 3 distance commands calculating `d12`, `d13`, and `d23`.
- Integer for loops now increment their final value, consistent with how for loops work everywhere else. Non-integer for loops retain previous behavior (final value of variable is last value of the loop).
- A new for loop, `datasetblocks` is introduced that allows looping over subsets (blocks) of fixed size of a data set,  or subsets of increasing size (cumulative). 
```
[help for datasetblocks]
	<var> datasetblocks <set> blocksize <#> [blockoffset <#>]
	[cumulative [firstblock <#>]]
```
For example, given a data set containing distances named `EndToEndAll`, one can loop over subsets of the data set in blocks of size 10 via:
```
for DS datasetblocks EndToEndAll blocksize 10 i=1;i++
```
The created subsets are regular data sets and each iteration can be referred to via the loop variable (`$DS` in this example). This will enable improved error analysis for users.
- `for` loop variables are stored in the master data set list as "string" variables. This makes the code involving "control blocks" like for loops easier and will facilitate adding new block types in the future (like `if`). I was able to remove several classes and consolidate functionality because of this. For loop subtypes have their own classes now, which makes things easier to read.
- Address #744.
- Direct copies of certain sets are now allowed via `=` (e.g. `set2 = set1`). If copy is not supported, the old behavior (renaming) is used.
- Online help on data set math now available via `help Math`.
- Vector data sets have now been split into "with origins" and "without origins". This makes a lot of the code under the hood simpler. What this means for output is that vectors with no origins (e.g. center of mass vectors from the `vector` command, etc) Will no longer have a "blank" origin (i.e. 0, 0, 0) printed. Internally this makes things like resizing and copying easier. This should probably be done for other data sets down the road (like covariance matrices etc).
- Manual updated and tests added.


drroe:
@hainm I've broken pytraj when I split DataSet_Vector into DataSet_Vector_XYZ and DataSet_Vector_OXYZ; a brief glance at e.g. c_dict.pyx makes me think this might not be an easy change. Can you take a look when you get a chance? I'd rather not revert the change since I think it's more sensible, but if it really breaks pytraj and its hard to fix I will consider reverting.

hainm:
hi @drroe, I will have a look as soon as I can. Yeah, I think it's not an easy fix since the vector dataset is used everywhere in `pytraj`.

swails:
Hai, the pytraj tests are very often hanging on Jenkins. Can you take a look at why?

hainm:
May be it’s due to multiprocessing issue in python 2 (pytraj has some tests
using it).  Can we switch to py3 on jenkin?

swails:
Are you sure it’s python 2? That doesn’t sound like something I would do.

hainm:
I will doublecheck. cheers.

hainm:
@swails I am right, jenkin is using python 2.7.

```
============================= test session starts ==============================
platform linux2 -- Python 2.7.15, pytest-4.6.9, py-1.8.1, pluggy-0.13.1 -- /opt/conda/bin/python
```

hainm:
@drroe Am I correct that the original `DataSet_Vector` is now called `DataSet_Vector_XYZ`?

drroe:
There are two new sets.

DataSet_Vector_XYZ is vectors with no origins. Essentially a single array
of Vec3.

DataSet_Vector_OXYZ is vectors and origins. Two separate arrays of Vec3.

drroe:
@hainm I've started to make changes to pytraj in this branch:
https://github.com/drroe/pytraj/tree/fix_vector_dataset
So far so good I think, but slow going since I'm not too familiar with the pytraj ecosystem.

hainm:
@drroe I just have a look and you're on the right track. thanks.

drroe:
I'm giving up on the vector data set stuff for now. I'll revisit it in the future I guess.

drroe:
Appveyor failure is because `ftp://ftp.unidata.ucar.edu/pub/netcdf/netcdf-4.3.3.zip` is no longer valid. Will update in a future PR.

PR #791, opened 2020-01-13
Fix detection of string data sets in standard data files.

drroe:
Previously strings like 105d would be converted to 105 with no errors and incorrectly detected as valid numbers - this fixes that.


drroe:
@swails note, I'm getting multiple Jenkins failures of the kind:
```
Error: docker: Error response from daemon: OCI runtime create failed: unable to retrieve OCI runtime error 
```

swails:
fixed.  Thanks!

PR #790, opened 2020-01-11
Make sure pytraj is tested with Python 3

swails:

PR #793, opened 2020-01-23
Trap PairList errors

drroe:
Errors that occur when setting up a PairList (due to e.g. corrupt coordinates) are now trapped. The total number of atoms off the grid are now reported instead of a message for every atom off the grid. Atoms that end up off the grid are now counted as problems by the `check` action. Pair list errors in energy calculations are now properly handled.

### DIFF
--- a/src/Action_Energy.cpp
+++ b/src/Action_Energy.cpp
@@ -416,6 +416,7 @@ double Action_Energy::Dbg_Direct(Frame const& frameIn, int maxpoints) {
 Action::RetType Action_Energy::DoAction(int frameNum, ActionFrame& frm) {
   time_total_.Start();
   double Etot = 0.0, ene, ene2;
+  int err = 0;
   typedef std::vector<CalcType>::const_iterator calc_it;
   for (calc_it calc = Ecalcs_.begin(); calc != Ecalcs_.end(); ++calc)
   {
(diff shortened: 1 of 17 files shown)


drroe:
It might be just about time to drop appveyor support. Failure is build system related and I'm tired of stamping out the unnecessary brushfires...

drroe:
@swails @hainm looks like pytraj tests timed out: https://jenkins.jasonswails.com/blue/organizations/jenkins/amber-github%2Fpytraj/detail/pytraj/127/pipeline

Any ideas why?

hainm:
I really don't know why. May be because of using python 2.7. 
I have disabled testing of `test_actionlist.py` (that might cause the hang). Please restart the build. thanks.

swails:
We stopped using python 2, I’m pretty sure. And that’s not a cause. “Removing an infinite hang” wasn’t one of the breaking changes they made in the migration to Python 3 ;).

hainm:
> We stopped using python 2, I’m pretty sure.

What do you mean? https://github.com/Amber-MD/cpptraj/pull/790

>  And that’s not a cause.

may be there's a memory leak in pytraj :D

drroe:
Any idea why we've got this pytraj hang? It shouldn't be related to these changes. I'd like to merge this but if there are any concerns I won't.

swails:
pytraj has been hanging for awhile. I suspect it’s some deadlock in one of the parallel tests.  It’s possible the problem lies in the mixed parallelism of the test framework with the tests themselves. The Jenkins nodes report having many CPUs while the Travis executors will report 1. I suspect the test framework will parallelize based on the number of available cpus, which would cause Jenkins to regularly hang while Travis wouldn’t. 

“Python 2” and “memory leak” wouldn’t cause a deadlock. I’d start by disabling the pytraj parallel tests. But I’ll defer to Hai on how to do that.

hainm:
>  I suspect it’s some deadlock in one of the parallel tests. 

seems reasonable to me too. 

> I’d start by disabling the pytraj parallel tests. But I’ll defer to Hai on how to do that.

I actually disabled the multiprocessing tests in Jenkins originally but forgot to turn off the MPI tests as well. I have made change here: https://github.com/Amber-MD/pytraj/blob/701c7adc86dae4b5c25924550295de76d164371b/Jenkinsfile#L33

hainm:
run jenkins

drroe:
@hainm Since pytraj compiles and tests fine with this branch on my local machine, I'm going to chalk the Jenkins failure up to something strange with the build machinery and merge this.

hainm:
yay.

PR #795, opened 2020-02-06
Add Amber Hub URL

drroe:
Ros Galindo over at U. Utah has come up with some nice CPPTRAJ (and Amber) info; link to it from README.


hainm:
look pretty.

PR #802, opened 2020-03-09
Fix Tinker ARC file detection when title line only has # atoms.

drroe:
The previous format check looked to see if the number of valid numbers on the tinker title line was the same as the number of tokens on the line; if so, detection would fail (because this may be another file format). This doesn't account for the case where there is only one number (a valid integer, the number of atoms), which is OK.


drroe:
Jenkins failure is that the Intel compilers are still not working there.

PR #806, opened 2020-03-26
Make 'nastruct resmap' more robust

drroe:
Previously `nastruct resmap` required all atoms in the reference to use exist in the residue to be mapped; for certain residues this is less than ideal. This PR makes it so that cpptraj will use whatever atoms from the specified reference it can find for the desired residue. As long as there are at least 3 atoms that can be RMS fit, the reference can be used.

This also adds a test for the `nastruct resmap` functionality.


drroe:
Jenkins failure is missing Intel compilers. Local build with Intel v 20 compilers passed all tests.

PR #807, opened 2020-04-02
Fixes for rotational diffusion analysis

drroe:
1) Change the default random seed to use the wallclock time if not specified instead of a hard-coded value. This was probably leftover from testing, but a hard-coded value should not be used as a default.

2) If the autocorrelation function of a random vector would integrate to less than 0, do not use that random vector for subsequent calculations (since it implies a negative effective diffusion constant for that vector). Manual updated with a note about this behavior.

### DIFF
--- a/doc/cpptraj.lyx
+++ b/doc/cpptraj.lyx
@@ -43357,11 +43357,8 @@ dt
 \series default
 ; the final time should be less than the total simulation time (see example
  below).
-\end_layout
-
-\begin_layout Standard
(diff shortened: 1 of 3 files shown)


drroe:
@dacase I probably should have asked you about this to be certain before I merged - does this seem reasonable? Skipping vectors whose autocorrelation function integrals end up being negative seemed like the best way to avoid NaN's in subsequent calculations.

PR #809, opened 2020-04-08
Add ability for 'datafilter' command to modify data sets.

drroe:
Previously, the `datafilter` command only generated sets with 1 for cutoffs satisfied and 0 otherwise. Now cpptraj has new keywords that allow one to modify a data set to only contain data satisfying the cutoffs:
```
filterset <set> [newset <newname>]
  If 'filterset' is specified, the specified <set> will be modified
  to only contain '1' frames; cannot be used with 'multi'. If 'newset'
  is also specified, a new set will be created containing the '1' frames instead.
```
This currently only works for scalar 1D sets. Test added and manual updated.


swails:
run jenkins

PR #810, opened 2020-04-21
Wrapdata

drroe:
Add new subcommand to `dataset` command, `shift`; can be used to apply offsets to data that match certain criteria:
```
shift [above <value> by <offset>] [below <value> by <offset>] <set arg0> ...
```
For example, to shift torsion data that is -180 to 180 to 0 to 360:
```
dataset shift TorsionSet below 0 by 360
```


drroe:
Not sure what happened with pytraj - builds for me on my local box.

PR #812, opened 2020-04-24
Add keywords to set default values for PDB occupancy/B-factor columns.

drroe:
For `trajout` etc. Allows users to specify something besides 1 or 0 for occupancy/B-factor columns respectively.
```
bfacdefault <#> : Default value to use in B-factor column (default 0).
occdefault <#>  : Default value to use in occupancy column (default 1).
```


drroe:
Since the only build I'm waiting for is CUDA, and this in no way impacts CUDA code, I'm merging.

PR #813, opened 2020-04-28
Add some simple minimization code.

drroe:
This code is not ready to be a full-fledged command (so the command will remain hidden), but I need it for some modelling I'm working on.

Adds the `emin` command and associated framework for doing simple steepest descent minimizations. This code was actually present in a different form for the cluster `drawgraph` functionality, but this makes it a bit more general. May find some general utility one day but for now it remains hidden and probably only useful to myself.


drroe:
Pytraj hung again - no idea why...

hainm:
I am not sure about `pytraj`. This one seems fine https://jenkins.jasonswails.com/blue/organizations/jenkins/amber-github%2Fpytraj/detail/pytraj/182/pipeline

PR #818, opened 2020-05-27
Use condaforge/miniforge3

hainm:
The hanging from pytraj test can be caused by either:
- the current image from anaconda
- openmp (default): I am testing this locally with `condaforge/miniforge3` (fine with non-openmp version).  --> Nope, it's fine with openmp for `condaforge/miniforge3` image.

### DIFF
--- a/devtools/ci/jenkins/Dockerfile.libcpptraj
+++ b/devtools/ci/jenkins/Dockerfile.libcpptraj
@@ -1,4 +1,4 @@
-FROM continuumio/miniconda3:latest
+FROM condaforge/miniforge3
 
 RUN mkdir /cpptraj && \
     mkdir /.conda /.local && \


hainm:
uhm, so it's `continuumio/miniconda3:latest` that caused the hanging. I have no idea why. lol.

swails:
the cause is an issue *somewhere* -- a combination of compiler options/libraries that causes a deadlock.  I'm not sure what's "different" about the miniconda3 image, but switching to miniforge3 is a reasonable workaround given that nobody has the bandwidth to find the real root cause.

@drroe - we should just go ahead and merge this.  That way we can actually start to get green Jenkins builds again :smiley:

drroe:
Thanks @hainm !

Issue #819, opened 2020-06-02
DSSP: Segmentation fault happens with bad pdb file


hainm:
This issue is originally from `pytraj`: https://github.com/Amber-MD/pytraj/issues/1443
and I did not have any clue why. I am trying today and it turns out (IMO) it's due to cpptraj itself. 

With on-disk trajectory (`trajin`), you need to run multiple times to reproduce the segmentation fault.

With on-memory trajectory `(loadcrd)`, the error happens all the times.

Please see the detail below:
- Using my macos, latest cpptraj commit: dab85feea4db5bb22683daa0b684ca60f80a7051

- Input
```
parm bad_pocket.pdb
trajin bad_pocket.pdb
dssp out abc  # Try to run 20 times, should get Segmentation fault in several runs


# uncomment to run
# loadcrd bad_pocket.pdb what
# crdaction what dssp out abc # Try to run 20 times, should get Segmentation fault in all runs
```

- pdb: 
[bad_pocket.pdb.txt](https://github.com/Amber-MD/cpptraj/files/4714718/bad_pocket.pdb.txt)

Cheers

hainm:
pytraj's user overcomes the segmentation fault by calling `set_new_mol` in pytraj (not sure if that's correct approach or not).

https://github.com/Amber-MD/pytraj/issues/1526#issuecomment-608411673

drroe:
Will be addressed by #820.

drroe:
Should be fixed now.

hainm:
confirmed. thanks Dan.

PR #820, opened 2020-06-02
Fix Topology::SoluteResidues() function when no molecule information present.

drroe:
If no molecule info present, original routine would segfault. Now try to use residue name to identify solute/solvent when no molecule info present. Addresses #819.


hainm:
image

Issue #825, opened 2020-06-30
GIST output file header


halx:
I am referring to cpptraj in Ambertools 20 with the first 4 patches applied.

There is a minor issue with the following line in Action_GIST.cpp

    datafile_->Printf("GIST Output, information printed per voxel\n"
                      "voxel xcoord ycoord zcoord population g_O g_H"
                      " dTStrans-dens(kcal/mol/A^3) dTStrans-norm(kcal/mol)"
                      " dTSorient-dens(kcal/mol/A^3) dTSorient-norm(kcal/mol)"
                      " dTSsix-dens(kcal/mol/A^3) dTSsix-norm (kcal/mol)"
                      " Esw-dens(kcal/mol/A^3) Esw-norm(kcal/mol)"
                      " Eww-dens(kcal/mol/A^3) Eww-norm-unref(kcal/mol)"
                      " Dipole_x-dens(D/A^3) Dipole_y-dens(D/A^3) Dipole_z-dens(D/A^3)"
                      " Dipole-dens(D/A^3) neighbor-dens(1/A^3) neighbor-norm order-norm\n");

There is an extra space in front of dTSsix-norm.  This can trip up file parsers badly.

drroe:
I see - it's after `dTSsix-norm`, before the `(kcal/mol)`. I'll work on a fix for GitHub. Patch for AmberTools will take longer as I don't have a lot of free time these days.

halx:
Thanks a lot.  No problem about the time.

If you have time though and think that would be useful too I would suggest to add some meta data to the output file, probably in the first line.  The idea here is to note down things like grid spacing/center/dimension etc. rather than have to supply that information to a post-processing script by hand.

drroe:
I don't suppose I could convince you to open a pull request implementing this could I? :smiley:

halx:
See #827. I have added a version number for the new format to distinguish from the old format and possible updates in the future.  Thanks.

drroe:
This seems fixed now. Please reopen if not.

PR #827, opened 2020-07-01
implementation for #825

halx:
This adds meta data to the GIST output file as discussed in #825.


drroe:
Looks like you just need to update the Test_GIST save files with the new header. Also, can you update the minor (second) version number in Version.h since the output has changed? Thanks very much!

drroe:
Pinging @gosldorf to make you aware of these upcoming changes. What do you think?

drroe:
Thanks - looks good!

PR #829, opened 2020-07-30
Misc. DSSP Fixes

drroe:
Some improvements to DSSP functionality:

1. Instead of limiting the number of bridges to a given residue to 2, make it unlimited.
2. Stop calculating N+1 to N hydrogen bonds (i.e. across a peptide bond). Doesn't seem to affect results, but does speed up assignment.
3. Improve the way previous/next residues are determined. Previously this was done by any bonds to other residues, but this isn't really the way the DSSP algorithm should work and would get confused by things like cross-links. Now make it so the previous residue is one bonded via the N atom and the next residue is one bonded via the C atom.
4. Residue setup is a bit more streamlined (lots of code cleanup).


drroe:
Jenkins has CUDA build failure related to driver/CUDA version incompatibility, and the PGI/Intel nodes are offline, so I'm going to merge this.

PR #831, opened 2020-08-11
Add ability to filter out alternate atom locations from PDB

drroe:
Adds new keyword to both `parm` and `trajin` commands for PDB files:
```
keepaltloc <char> : If specified, alternate location ID to keep.
```
Allows one to keep only a specific alternate location ID if desired. Update manual and add a test.


drroe:
Jenkins failure is that no CUDA-capable devices were found on the runner, so not a code failure.

PR #832, opened 2020-08-12
Fix CUDA builds

swails:
The new CUDA agents on Jenkins use a different nvidia runtime for docker that requires the GPUs made available to the container be declared at the command-line.  This change implements that behavior and will fix CUDA builds from here on.


drroe:
Thanks @swails! Pytraj error should now be fixed by https://github.com/Amber-MD/pytraj/pull/1547.

PR #835, opened 2020-08-18
Add 'pdborder' keyword to 'fixatomorder' command

drroe:
Allow topology to be reordered by PDB info. Still testing so it will remain "hidden" (undocumented) for now.

Also, the Append() function of Topology now preserves the original residue numbering field.


swails:
You are a broken record, @AmberJenkins

hainm:
I don't know who is the bot anymore. lol.

PR #837, opened 2020-08-21
Add experimental command 'fixmissingresidues'; other miscellaneous improvements.

drroe:
This PR covers a lot of ground; some of it is still pretty alpha (beta at best), but there are enough fixes now that I want to make sure they get merged in.

**The improvements:**

- Allows copying of COORDS sets, e.g.
```
SET2 = SET1
```
Will now create `SET2` as an independent copy of DataSet_Coords_CRD set `SET1`.
- Allow negative numbers in original residue number (`:;`) mask expressions, since PDB residue numbering allows negative numbers.

**New functionality:**
- Add `auto` keyword to the `box` command:
```
auto [offset <offset>] [radii {vdw|gb|parse|none}]
```
This will create an orthogonal bounding box using the specified radii and offset. Useful if e.g. you want to use pairlist code but no box exists (for something like `check`).

**Experimental functionality:**
- Add keywords to `fixatomorder` to try to put atoms in PDB order:
```
[pdborder [hetatm <mask>]]
```
Attempts to organize atoms by PDB information (i.e. Chain ID, original residue numbering, and insertion codes). Atoms optionally specified by 'hetatm <mask>' will be placed after all other atoms. Seems to work reasonably well but definitely needs more testing, especially when it comes to insertion codes.
- Add `addmissingres` command. This attempts to take a structure with missing residues (a PDB file) and model them in. It is very, very, VERY simplistic and still needs a lot of work, so it is hidden for now. I won't even post the command syntax here because it will surely change.
- Add `sugarmask` and `determinemolmask` keywords to the still hidden `prepareforleap` command. The `sugarmask` keyword will attempt to identify sugars, change them to GLYCAM names, unbond them, and create the necessary LEaP input for re-attaching them in LEaP (like disulfides). The `determinemolmask` keyword is for parts of the structure where bonding should be used to determine where chains end (in case things are missing from chains so we don't put TER cards we dont want).

Also adds tests and updates the manual for non-hidden stuff.

### DIFF
--- a/test/Test_Box/RunTest.sh
+++ b/test/Test_Box/RunTest.sh
@@ -10,6 +10,8 @@ TESTNAME='Box tests'
 Requires netcdf maxthreads 10
 
 INPUT="-i box.in"
+
+UNITNAME='Box Test (Add box info)'
 cat > box.in <<EOF
 parm ../tz2.parm7
 trajin ../tz2.nc 1 10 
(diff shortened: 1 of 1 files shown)


drroe:
> It's a real failure. It just takes a bunch of clicks to find it.

Indeed, I should have dug deeper. Thanks for pointing me in the right direction!

PR #838, opened 2020-08-21
Fixes and improvements for SPAM and volmap, 3d grids

drroe:
This PR has multiple fixes and improvements related to SPAM, volmap and reading/writing 3D data.

- The online help and manual documentation are greatly improved for both.
- The `volmap` command now creates a data set for the peaks data. This means that among other things you can run `spam` and `volmap` in the same cpptraj run. This could include running volmap several times with the same peak data and different settings.
- Some checks have been introduced into `spam` to trap "bad" peaks (poor occupancy, extremely large energies, etc) that previously could lead to the code producing nonsense or failing outright.
- OpenDX data can now be read in single (the previous behavior) or double precision if desired.
- OpenDX output now respects if the format of a data set has been changed (via e.g. `dataset precision`) from the default of `%g`.
- Enable divide and multiply for 3D datasets.

Miscellaneous fixes:
- Improve Tinker format recognition.

### DIFF
--- a/src/Version.h
+++ b/src/Version.h
@@ -12,7 +12,7 @@
  * Whenever a number that precedes <revision> is incremented, all subsequent
  * numbers should be reset to 0.
  */
-#define CPPTRAJ_INTERNAL_VERSION "V4.29.4"
+#define CPPTRAJ_INTERNAL_VERSION "V4.29.5"
 /// PYTRAJ relies on this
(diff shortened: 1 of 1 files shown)


drroe:
> It actually didn't pass. 

@swails you're 100% right! Thanks - I think I've found the issue, should be fixed by #839 (I hope).

PR #839, opened 2020-08-24
Fix IRED for PGI compilers

drroe:
Some iterator math in an inner loop of the IRED analysis seems to have confused PGI compilers. This PR slightly refactors the iterator usage to eliminate the issue.


swails:
Well the Jenkins build passed without posting a message about the PGI compilers, so it looks like that did the trick!

PR #484, opened 2017-04-09
Use exact (and stable) version: pytraj-build-box:17.0

hainm:

PR #840, opened 2020-08-24
Better control over printing of CONECT records to PDB files.

drroe:
Previously, CONECT records were printed for all bonds if `conect` was specified, or only HETATM residues if `pdbres`/`pdbv3` had been specified. This PR adds a new keyword, `conectmode`, which allows more direct control over the printing of CONECT records:
```
conectmode <m>  : Write CONECT records for <m>='all' (all bonds), 'het' (HETATM only), 'none' (no CONECT).
```

Manual updated and test added.


drroe:
@swails So the PGI failure seems like the old machine-specific one. I notice that the Jenkinsfile changed from PGI requiring  `batwoman` to `intel-cpu` - can we change it back or is `batwoman` gone? It seems like the one that passed in fact ran on [Batwoman](https://jenkins.jasonswails.com/blue/rest/organizations/jenkins/pipelines/amber-github/pipelines/cpptraj/branches/PR-839/runs/1/log/?start=0), while the failed one seems to have run on [Green Lantern](https://jenkins.jasonswails.com/blue/rest/organizations/jenkins/pipelines/amber-github/pipelines/cpptraj/branches/PR-840/runs/2/log/?start=0).

Also, the Pytraj failure can probably be fixed by sourcing the `cpptraj.sh` script (to set CPPTRAJHOME so the J-coupling action can find the data file it needs) that is generated by cpptraj's install. Is there an easy way to do that? I looked at the Jenkins files, but I think all that logic is in the docker images right?

All that said, I've independently verified that everything builds and tests with PGI, so I think this is safe to merge.

swails:
Batwoman is a very old machine (ca. 2011) with a Xeon processor.  The other Intel CPU machines all have modern i5s (3 machines with 9400F).  I'm a little wary of tying the PGI build to the oldest machine in the array since if it goes down, no build can finish.

When the machine-specific issue first arose, the two machines in the Jenkins array was an AMD FX-6100 (ca 2012) and a Xeon X5675 (ca 2011), and it only worked on the Xeon.

Give me a few days to dig a little and see what I can find.  Alternatively, I can grant you ssh access to Green Lantern if you wanted to dig, as well.

drroe:
@swails I think this warrants its own issue: see #484 

I've been able to reproduce the failure at CHPC Utah; helpfully they have both an architecture that PGI-compiled cpptraj passes for, as well as one it fails for. I'll continue debugging there for the moment.

PR #859, opened 2020-11-09
Add particle mesh Ewald and long-range LJ correction to GIST energy calculation

EricChen521:
Hi Dan,

This is the code we used in the PME-GIST manuscript. Although we are still working on some additional functionalities in GIST, we'd better pull the request now so that the reviewer can see our code is already open source.

Please let me know if you have any suggestions regarding the code.

Eric


drroe:
Sorry for the delay in replying - I'm on leave so my time to work is very limited.

> Although we are still working on some additional functionalities in GIST,

We can keep it as a work in progress until the issues are resolved. This way at least it's here in some form. I'll try to take a closer look at the code itself over the next few days.

src/Action_GIST.h
```diff
@@ -6,6 +6,17 @@
 #ifdef CUDA
 #include "cuda_kernels/GistCudaSetup.cuh"
 #endif
+
+#include "DataSet_3D.h"
```
drroe, line 10: We may want to see if some of these includes can be replaced with forward declares.

src/Ewald_ParticleMesh.cpp
```diff
@@ -265,7 +344,175 @@ int Ewald_ParticleMesh::CalcNonbondEnergy(Frame const& frameIn, AtomMask const&
...
+
+//  MapCoords(frameIn, ucell, recip, maskIn);
+  double e_recip = Recip_ParticleMesh_GIST( frameIn.BoxCrd(), e_potentialD );
+
+  auto step2 = std::chrono::system_clock::now();
+
+
+  //mprintf("e_recip finished, the e_recip value: %f \n", e_recip);
+
+  std::chrono::duration<double> d2 = step2 - step1;
```
drroe, line 432: Cpptraj has a Timer class for timing.

src/Ewald_ParticleMesh.cpp
```diff
@@ -265,7 +344,175 @@ int Ewald_ParticleMesh::CalcNonbondEnergy(Frame const& frameIn, AtomMask const&
...
+  t_total_.Start();
+  Matrix_3x3 ucell, recip;
+  double volume = frameIn.BoxCrd().ToRecip(ucell, recip);
+
+  auto step0 = std::chrono::system_clock::now();
+
+
+  double e_self = Self_GIST( volume, e_elec_self ); // decomposed E_elec_self for GIST
+
+  auto step1 = std::chrono::system_clock::now();
```
drroe, line 380: Because cpptraj is distributed with Amber, C++11 stuff still has to be optional. All C++11 stuff should be hidden behind the define `C11_SUPPORT`

drroe:
Superseded by #889

Issue #861, opened 2020-11-24
It's time to abandon Travis-CI and move to GitHub Actions (or entirely Jenkins)


swails:
GHA will take some time, but has mac builds.  Jenkins is effectively already supported, and once the Utah machine comes online will be more reliable.

First step is to probably just turn off Travis integration.

They're basically ending support for free Open Source builds - https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works

drroe:
What a bummer. Well, it was fun while it lasted. I had noticed the Mac builds were "failing" out of the blue but hadn't had time to look into it. While I don't begrudge Travis CI the right to charge for services, it seems like they could have gone about this transition a better way. 

Either option (all Jenkins or GA) is fine with me. I'll have to educate myself on GA, not sure how much work will be involved, although from the post you linked to it seems entirely doable. I'm not fully back until next week but I'll try to make some small progress (e.g. disabling Travis) until then.

swails:
I'll whip up a GHA pipeline at some point when I get a chance (I've been working with them in my new job).  Under Microsoft's stewardship I expect GHA to be free for OSS repos.

We could alternatively use Azure pipelines, but GHA is easier to set up and manage.

swails:
Cpptraj is on GitHub actions and off Travis

PR #862, opened 2020-12-06
Have the 'checkstructure' action respect exclusions

drroe:
Currently only bond exclusions since only abnormal bond lengths are check. Also check for abnormally short bonds, with offset controlled by new keyword, `minoffset`. Exclusions now handled by a new separate class, ExclusionArray, instead of being stored in Atom and set up by Topology; this allows much more flexibility over how exlclusion lists are created and used. Performance for `check` with no PBC (i.e. no pair list) is pretty much unchanged; performance for check with the pair list suffers about 5%; this can't be helped and is entirely due to the fact that the exclusion array needs to be checked. Probably some tweaks could be done since the exclusions are only bonds so far, but the current code will work for any type of exclusion.

At some point I should come back and optimize all the non-pair list loop code to be like how StructureCheck is optimized.


drroe:
Appveyor failure is not code related; problem downloading mingw:
```
:: Retrieving packages...
downloading mingw-w64-x86_64-c-ares-1.15.0-1-any.pkg.tar.xz...
error: failed retrieving file 'mingw-w64-x86_64-brotli-1.0.7-1-any.pkg.tar.xz' from repo.msys2.org : The requested URL returned error: 404
```

drroe:
The Travis OSX builds are effectively no longer supported, related to #861 (and https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing). I'm going to remove the "required" status from Travis and move it to Jenkins @swails.

PR #863, opened 2020-12-07
First pass at a GHA workflow for cpptraj

swails:


swails:
I think the first iteration needs to come from this repo itself.

PR #865, opened 2020-12-09
Fix 'molinfo short'

drroe:
Fixes a potential segfault when a residue name mismatch occurs and the residue size is 1.


drroe:
Appveyor failure is the mingw download issue again. Not clear to me whether Travis will even start - might just merge since Jenkins thinks it's cool (other than PGI, but that's a whole different issue)...

PR #843, opened 2020-09-02
Add support for CMAP in latest AMBER force fields.

HanatoK:
Solve issue #842 myself

PR #872, opened 2021-01-22
Fix errors reading some Amber topology files with CMAP parameters

drroe:
The CMAP reading code introduced in #843 assumed that if the `CTITLE` flag was present, the CMAP flags would have the style `CHARMM_CMAP_XXX`; however, this is not always the case (e.g. for topologies produced by CHARMM GUI) and cpptraj would fail reading the topology in such a case. This MR rewrites the CMAP parameter reading routines so they are more general and accept either `CHARMM_CMAP_XXX` or `CMAP_XXX` style no matter if `CTITLE` is present or not.

### DIFF
--- a/src/Parm_Amber.cpp
+++ b/src/Parm_Amber.cpp
@@ -413,7 +413,8 @@ int Parm_Amber::ReadNewParm(Topology& TopIn) {
           ptr = SkipToNextFlag();
         } else {
           int err = 0;
-          switch ((FlagType)flagIdx) {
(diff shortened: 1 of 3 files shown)


drroe:
Appveyor failure is the mingw stuff that has been broken for a while. Travis is taking forever (need to migrate). As soon as LGTM says things are good I will merge.

swails:
I have a pr opened to migrate to GHA. I just have to finish it. Some of the builds are working.

PR #873, opened 2021-01-22
Try to fix appveyor builds.

drroe:
Try to fix mingw build issues by upgrading to windows server 2019.


drroe:
Looks like switching to VS2019 image fixed it!

Issue #841, opened 2020-08-24
CPU-dependent Test failures with PGI compilers


drroe:
Some original discussion was in #840 

Parts or all of the following tests fail when compiled with PGI compilers, but only for certain types of hardware:
```
Test_Cluster_RandomSieve/Test_Error.dat
Test_Cluster_Sieve/Test_Error.dat
Test_Cluster_SymmRMSD/Test_Error.dat
Test_CurveFit/Test_Error.dat
Test_EvalPlateau/Test_Error.dat
Test_GIST/Test_Error.dat
```

Failures occur on the following platforms:
- Intel(R) Xeon(R) Gold 6130 CPU @ 2.10GHz

Failures do not occur on the following platforms:
- Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz

It's possible the failures are optimization-related; e.g. the Intel 6130 supports AVX-512 while the E5-2670 does not. I will try compiling with no optimizations and see if the test failures go away.

drroe:
Even with `-O0` the tests still fail.

drroe:
Using flags recommended by [Blue Waters](https://bluewaters.ncsa.illinois.edu/pgi-compiler):
```
-m64 -mcmodel=medium -Mdalign -Mllalign -Munroll -Kieee -O2 -fastsse -Mipa=fast
```
results in the same test failures, plus `Test_CCP4/Test_Error.dat`.

drroe:
Finally addressed by #876 !

PR #876, opened 2021-01-30
Fix test differences with PGI compilers

drroe:
This MR at long last should address #841. It includes some small fixes and reworks some tests so that the code will return nearly identical results to GNU/Intel compilers.

- The best-representative code in clustering was reworked so that 1 and 2 member clusters are handled in a more consistent manner.
- The symmetric RMSD clustering test uses a more realistic trajectory. The artificial trajectory had too many identical frames, and the zeros in the pairwise matrix could lead to unpredictable behavior.
- In curve fitting, one loop in the Householder transformations was forced to de-optimize, which seems to bring the PGI results more in line with GNU/Intel.
- Tolerances were relaxed slightly for the `evalplateau` test.
- In the GIST test, a vector sign check was made more numerically stable by comparing to `Constants::SMALL` instead of zero.


drroe:
@swails from Jenkins:
```
All nodes of label ‘linux’ are offline
```
Will they be back up soon?

drroe:
Jenkins is back! Yay!

drroe:
PGI build passed!

swails:
The Jenkins nodes actually shut down and reboot (via wake on lan) as they are needed.

Doing that I was able to cut the power usage from 20-25 kWH per week to ~5 kWH per week. 

It's actually a pretty nifty bit of code if I do say so myself, as was the setup to actually monitor the usage. 

But what that means is that it may take about 2-3 minutes after a job is submitted before an agent comes online to execute it.

swails:
@drroe do you want me to make pgi failures fatal again now that they should be passing?

drroe:
> The Jenkins nodes actually shut down and reboot (via wake on lan) as they are needed.

That is awesome!

> do you want me to make pgi failures fatal again now that they should be passing?

The lazy person in me wants to say no, but since I have access to the compilers now and can test them more regularly I guess PGI errors should be fatal. I just wish I had a sense of whether they're still worth actively supporting or not. I haven't used them in a production capacity since I was at the University of Utah (so < 2016).

swails:
The most compelling argument to me is that using multiple compilers increases the likelihood of finding subtle bugs.  Obviously this argument falls apart when most of what you uncover are compiler bugs.

My take here is that the core issues were as much problematic tests (susceptible to numerical instability) as flaky compilers.  If the net result of fixing PGI compatibility was the introduction of more realistic test cases with more robust and well-defined behavior, then I'd argue that's a good thing.  De-optimizing a loop to work around a compiler bug is a bad thing.  If the latter is minor, then I think gating on PGI passing is good, but I defer to you.

PR #877, opened 2021-01-31
Add ability to read matrix in "XYZ" format; new 'flatten' command

drroe:
Non-square matrices can now be read in via the `read2d` keyword, i.e. matrix data in the format:
```
<column> <row> <data>
```
The `nosquare2d` keyword needs to be specified.  I should probably try to do a better job autodetecting this in the future.

Also adds the `flatten` command, used to convert a matrix to 1D array by dividing matrix elements up among the array somehow.
```
> help flatten
  [help flatten]
	name <output set name> [mode {sum|avg}] <input set args>
```
Useful for e.g. distributing elements of a pairwise matrix across its elements. For example, given a matrix with values like this:
```
X Y Value
1 3 5.0
1 4 4.0
2 3 2.0
```
The “flattened” 1D array with mode SUM would be determined as follows:
```
Element 1 = (5.0/2) + (4.0/2) = 4.5
Element 2 = (2.0/2) = 1.0
Element 3 = (5.0/2) + (2.0/2) = 3.5
Element 4 = (4.0/2) = 2.0
```
And the final 1D array would look like so:
```
Index Value
1     4.5
2     1.0
3     3.5
4     2.0
```


drroe:
I'm getting messages like `Failed to extract source.tar.gz` in post-test steps. Maybe I'll try again tomorow.

swails:
> `Failed to extract source.tar.gz`

That's a transient Jenkins error.  It was stashed in the first step and should be available in every future step.  Re-running should fix it.

drroe:
I'm not sure why, but Jenkins seems unstable lately. One run (number 5) took ~14 hours to fail; problem was a node unexpectedly closing down:
```
Cannot contact Supergirl: hudson.remoting.ChannelClosedException: Channel "hudson.remoting.Channel@34364954:Supergirl": Remote call on Supergirl failed. The channel is closing down or has closed down
```
The following restarted run failed on the Intel step because it couldn't extract the source:
```
Failed to extract source.tar.gz
```
I guess I will try restarting, but I eventually may need to remove the required status from Jenkins just so I can get this merged.

swails:
> I guess I will try restarting, but I eventually may need to remove the required status from Jenkins just so I can get this merged.

I think I found the issue with premature shutdowns and I've fixed it.  I think switching the required builds to several of the GHA builds is a good move - I'd recommend switching regardless (I don't think Jenkins does any CMake builds, for instance).

Requiring the CMake builds that take 11 and 15 minutes will test a wider variety of options (OpenMP, MPI, sander API, etc.).

PR #882, opened 2021-02-26
Add new random number generators

drroe:
The current Marsaglia generator that cpptraj uses (the Amber default one) is known to have issues. This PR is the first step to adding better random number generators. See `devtools/rngtest/README.md` and `devtools/testu01/README.md` for test results using the Dieharder and TestU01 RNG test suites respectively.

The RNGs added are:

1. The C-standard library RNG.
2. The Mersenne Twister (via C++11 library)
3. Permuted Congruential Generator 32 (https://www.pcg-random.org/)
4. Xoshiro128++ (http://prng.di.unimi.it/)

Numbers 2, 3, and 4 are the best so far, and I think eventually 4 may be made the default based on initial test results.

@hainm This also adds `ExternalFxn.h` ; the idea is eventually to get everything that packages like pytraj need (but not necessarily cpptraj itself) in there, and hopefully I will be less likely to break interfaces to those packages. I will be making a PR to pytraj soon that will make use of this and will protect the pytraj tests when the default RNG is changed.

This PR also adds the manual entry for `evalplateau` (since it is now published, https://aip.scitation.org/doi/abs/10.1063/5.0013849), and some missing citations. Also, the Travis badge is swapped for GitHub Actions.


hainm:
> @hainm This also adds `ExternalFxn.h` ; the idea is eventually to get everything that packages like pytraj need (but not necessarily cpptraj itself) in there, and hopefully I will be less likely to break interfaces to those packages. I will be making a PR to pytraj soon that will make use of this and will protect the pytraj tests when the default RNG is changed.

image

drroe:
@swails Thanks for the help squashing that `tzdata` issue!

PR #884, opened 2021-03-04
Add location of CPPTRAJ source directory to cpptraj.sh

drroe:
Makes it easier for the separate repository of extra tests to find critical test files.

Also add a check to `configure` to prevent it from being run from somewhere other than the source directory, which is currently unsupported; I should maybe change that at some point.


drroe:
@swails I've been having trouble checking Jenkins today - lots of timeouts.

drroe:
Going to merge since everything else has passed.

PR #885, opened 2021-03-05
Reorganize test utilities, add unit tests.

drroe:
Reorganization of test utilities to hopefully make things simpler to maintain. Instead of being in a directory separate from `test`, they are now in a subdirectory of `test`.

- Move `ndiff.awk` to a `utilities` subdirectory of the `test` directory; it is only used by tests and really belongs there.
- Delete the `nproc` program with a (hopefully) simpler to use and maintain shell script that determines number of processes by counting how many `echo`s the current `$DO_PARALLEL` will invoke. Also have this in `test/utilities`.
- Add unit tests that were in an external repository here. The idea is that they test very basic units of functionality and are either pass/fail. Try to make sure I don't break fundamental functionality in e.g. the `ArgList` class. These will be tested on Jenkins.


drroe:
Grr - I hate when CI fails due to the infrastructure, although I guess you get what you pay for...
```
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/n/nvidia-settings/libxnvctrl0_440.82-0ubuntu0.20.04.1_amd64.deb  404  Not Found [IP: 52.147.219.192 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

drroe:
Looks like adding apt update to Actions did the trick. LGTM failed due to timeout but that seems to be a problem on their end.

drroe:
I don't know why LGTM is failing. It just keeps timing out. The last step is this:
```
[148/150 eval 56ms] Evaluation done; writing results to codeql-cpp/Metrics/Files/FLinesOfComments.bqrs.
Starting evaluation of codeql-cpp/Metrics/Files/FNumberOfTests.ql.
```
There are no error messages as far as I can tell. I'm tired of debugging it. If Jenkins passes again I'm just going to merge.

PR #886, opened 2021-03-10
Add a timeout to Jenkins build pipeline

swails:
This will prevent zombie jobs from running amok.  It actually filled the Amber CI machine with 1.4 TB of logs last week... this change will fix that.


drroe:
Thanks!

PR #887, opened 2021-03-15
Fixes and improvements related to random number generation. Improvements for data set filtering.

drroe:
This PR contains a lot of improvements.

- Random number generation is greatly improved under the hood, and after some more testing the default random number generator should be changed (lead candidate right now is Xoshiro128++). Added a `gauss` keyword to the `random` command.
- The algorithm for the `permutedihedrals` keyword with the `random` command is greatly improved. In general, fewer rotations are required to get much higher quality structures (with no clashes so far in testing proteins up to 268 a.a.!).
- The algorithm for `randomizeions` has been sped up an order of magnitude. In addition, users are not required to specify an `around` mask, and in conjunction with `allowoverlaps` eliminates all distance checking, allowing for pure random ion placement.
- The code for `datafilter` and the `filter` action has been centralized under the hood. In addition, the final counts of passed/filtered frames are now stored in data sets, and 2d/3d sets can be counted (but not yet actively filtered).

My hope is that this is the version (5.1) that ends up in AmberTools 21.


drroe:
FML
```
E: The repository 'https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_20.04  Release' no longer has a Release file.
```
I just want stable CI....

swails:
What package is it trying to install?

.github/workflows/merge-gate.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 jobs:
   linux_build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
swails, line 10: Same image ATM.

drroe:
@swails its failing on the `apt update` command, which I had to add previously because I was getting "Unable to fetch some archives" messages without it (see https://github.com/Amber-MD/cpptraj/pull/885).

drroe:
@hainm note the pytraj failure is expected; the algorithm for the `randomizeions` command has been improved, but now produces different output. The `originalalgorithm` keyword can be used to recover original behavior, or the test output will need to be updated. I can do that after I merge this.

drroe:
@swails taking out `apt update` seems to have fixed it this time. I guess actions is still smoothing out the rough edges.

PR #894, opened 2021-04-16
Support AMD GPUs (using HIP)

ex-rzr:
I cherry-picked commits from AmberTools/src/cpptraj but I think they can be squashed because there not many changes overall.

Building instructions are similar to instructions for CUDA:
```
export HIP_HOME=/opt/rocm
./configure --buildlibs -hip gnu
```


drroe:
@ex-rzr Thanks for this - looks promising!

@jokr91 I'm pinging you since there are some code changes to PME GIST you may want to review. They seem mostly innocuous, but I wonder if e.g. there is a reason you used things like `*= -1` instead of the `abs()` function in `GistCudaCalc.cu`, etc.

jokr91:
@drroe as far as I remember they are very similar to the dist functions defined in DistRoutines.cpp. I did not want to change away from the definitions there and thus there is the `*= -1` in there (though I think in DistRoutines.cpp it is `x = -x`). However, there was a reason I did not use floats but doubles at some points in the code, because this sped up the calculatiuon on my GPU (I think this had something to do with double being calculated on different multiprocessors than float, but I am not sure on that right now and it might be different on various platforms).

drroe:
@ex-rzr I'm merging the latest changes from master into here, should fix the Jenkins build - hopefully that's ok.

ex-rzr:
> @ex-rzr I'm merging the latest changes from master into here, should fix the Jenkins build - hopefully that's ok.

Thanks. I forgot to merge the current master, I used the last commit of cpptraj "merged" into AmberTools/src/cpptraj just for convenience sake so I wouldn't change something incorrectly or forget to change something else.

drroe:
@ex-rzr I see that this MR includes a GitLab CI file. Is this for a local repo or is it for somewhere I can add a hook to? Since there is no AMD testing in the current CI setup it would be great to add it to ensure this code is protected against people like me messing it up.

configure
```diff
@@ -19,6 +19,7 @@ UsageSimple() {
   echo "    -mpi           : Use mpicc/mpicxx to compile."
   echo "    -intelmpi      : Use mpiicc/mpiicpc to compile."
   echo "    -cuda          : Enable CUDA support. See SHADER_MODEL in --full-help for more info."
+  echo "    -hip           : Enable HIP support. See SHADER_MODEL in --full-help for more info."
```
swails, line 22: SHADER_MODEL only references CUDA -- can you update the description to include its relevance to HIP/AMD?

ex-rzr: Done. There are also a few more small changes, please check.

ex-rzr:
> @ex-rzr I see that this MR includes a GitLab CI file. Is this for a local repo or is it for somewhere I can add a hook to? Since there is no AMD testing in the current CI setup it would be great to add it to ensure this code is protected against people like me messing it up.

@drroe 
It's for our private repository. It seems that it is possible to use GitLab CI as a check for GitHub:
https://docs.gitlab.com/ee/ci/ci_cd_for_external_repos/github_integration.html
https://docs.gitlab.com/ee/user/project/integrations/github.html

According to the second page the check should work without complex configuration. I've tried it on my fork of cpptraj but I'm not able to make GitHub run GitLab CI (mirroring works though), I think some additional configuration in Actions is required.
So I need help with this as I don't have experience with GitHub Actions.

drroe:
> I've tried it on my fork of cpptraj but I'm not able to make GitHub run GitLab CI (mirroring works though),

That's fine. My concern is just keeping the code healthy. If you will be running some tests locally that can work as a stop-gap until I figure out some way to get AMD CI going. I'm going to try to get some AMD GPUs to play with here at the NIH (if we have the budget). Otherwise I might just try to get a Radeon to throw in my own Linux box.

@swails any other concerns? If not I think this is good to merge. Thanks @ex-rzr !

PR #896, opened 2021-04-23
Improvements and Enhancements for GIST PME. Add 'topoint' keyword to 'translate' action.

drroe:
Version 5.3.0.

Several improvements and enhancements for GIST:

- The energy calculation with PME has been sped up about 40% by ignoring all nonbonded interaction completely off the grid. Total energy is no longer reported (it was never really needed except for debugging).
- The Eww, Esw, and neighbor grids are now determined when using PME.
- Update the citation for GIST PME (paper is now published): https://pubs.acs.org/doi/10.1021/acs.jctc.0c01185
- Add `neighborcut` keyword to allow customization of the neighbor cutoff (default is still 3.5 Ang.).
- Add `floatfmt`, `floatwidth`, `floatprec` keywords for changing format, width, and precision of floating point values printed to the GIST output file. Results for `pme` no longer automatically have a different output format.
- Add `intwidth` keyword for changing width of integer values printed to GIST output file.
- Reduce the size of the PME GIST test files.

This also adds a new keyword to the `translate` action, `topoint` (to point), which allows translating the center of the atoms specified in the given atom mask to a specific set of coordinates, e.g. `translate :1 topoint 12.5,12.5,12.5`. (This was very useful for testing GIST by moving waters into specific voxels).

Manual updated, tests updated and added.


drroe:
Jenkins timed out, seems like it took a while for the PGI build to get going. I'm going to test PGI locally, if it passes I will merge.

drroe:
PGI 17.10 and Intel 19.0.5 passed. Going to merge.

PR #898, opened 2021-04-28
Some GPU code cleanup

drroe:
Add header files for `core_kernels.cu` and `kernel_wrappers.cu`, and fix up the dependencies. Do some minor cleanup, get rid of some `nvcc` warnings.

Also remove the `VERSION` file from `xdrfile` and put that info in `README`. On some versions of XCODE (OSX) having a file named `VERSION` can get mistakenly pulled into `cstddef` leading to compilation failure.


drroe:
@ex-rzr I don't think I broke anything AMD-related, but if you have a free moment and can test I would appreciate it. Thanks!

drroe:
@swails Several Jenkins builds are hanging on the `source` stage. Any idea why?

swails:
CUDA tests passed.  This looks like connection issues between the Jenkins master and the Utah machine.  I'm updating and rebooting that agent which usually corrects these problems.

In the meantime, enough tests passed that I think this can just be merged.

ex-rzr:
@drroe I will run tests tomorrow (but at first glance it looks ok)

ex-rzr:
I've just checked on AMD: everything is ok

drroe:
@ex-rzr Thanks!

PR #903, opened 2021-05-17
Add basic support for Intel OneAPI

drroe:
Adds a `oneapi` target for `configure` to use the `icx`, `icpx`, and `ifx` compilers. I haven't done any performance testing yet - the purpose of this PR so that such testing can be done. I've tested that cpptraj will compile with all compilers from OneAPI (`icpx`/`icpc`), and that the `icpx`/`gfortran` combo works if the OneAPI HPC add-on is not installed.

Also adds a missing include to the bundled `readline` code that seems to be causing issues with OSX/clang12.

Since no cpptraj code is changed, no version bump.

### DIFF
--- a/configure
+++ b/configure
@@ -1329,7 +1329,7 @@ SetupCompilers() {
   if [ ! -z "$CC"  ] ; then echo "C compiler (CC) set to $CC" ; fi
   if [ ! -z "$FC"  ] ; then echo "Fortran compiler (FC) set to $FC" ; fi
   # If no compiler type specified try to guess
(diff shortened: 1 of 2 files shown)


drroe:
Jenkins failures are framework-related, not cpptraj:
```
[2021-05-17T18:31:43.337Z] Error response from daemon: Get https://registry-1.docker.io/v2/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)

script returned exit code 1
```

PR #905, opened 2021-06-14
Fix `configure` behavior for "optional" libraries

drroe:
Fixes behavior of `configure` for parallel NetCDF and FFTW.

Parallel NetCDF is now considered optional for MPI (instead of required); this will fix compile behavior in AmberTools when parallel NetCDF is not available (since under AmberTools checks are disabled).

FFTW is now enabled by default but allowed to be optional (since FFTW is required for PME functionality). Will still use FFTW from MKL if `-mkl` has been specified and `-fftw3` has not.

This PR also fixes the GitHub Actions tests. Previously, parallel tests were not actually being run (`DO_PARALLEL` was not set). Also, the cmake build was building separate MPI and OpenMP binaries, not a hybrid MPI+OpenMP binary. Switched the OpenMP, MPI, and MPI+OpenMP tests to use the `configure` build instead of cmake.

### DIFF
--- a/.github/workflows/merge-gate.yml
+++ b/.github/workflows/merge-gate.yml
@@ -20,15 +20,15 @@ jobs:
           - use_openmp: "yes"
             compiler: "gnu"
             use_mpi: "no"
-            use_cmake: "1"
+            use_cmake: "0"
           - use_openmp: "no"
             compiler: "gnu"
             use_mpi: "yes"
-            use_cmake: "1"
+            use_cmake: "0"
           - use_openmp: "yes"
             compiler: "gnu"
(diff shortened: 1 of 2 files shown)


drroe:
```
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/n/nvidia-settings/libxnvctrl0_460.39-0ubuntu0.20.04.1_amd64.deb  404  Not Found [IP: 52.147.219.192 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
Error: Process completed with exit code 100.
```
CI framework failures do not spark joy...

Issue #914, opened 2021-09-01
Determine representative frames for combined clustering


drroe:
Currently cpptraj only reports best representative frame(s) for entire clusters. When combined clustering is performed, may want to also have the best representative(s) within the splits.

drroe:
Addressed by #915

Issue #805, opened 2020-03-25
DBSCAN clustering should provide a better error message for mutually exclusive options


drroe:
Specifically `kdist` and everything else.

drroe:
Addressed by #915

Issue #769, opened 2019-10-22
Should be able to cluster on existing pairwise distance matrices


drroe:
Currently you need to specify some sort of data set to cluster on, but it's not strictly necessary. Issue was raised by user Tim Davis on the Amber mailing list. Here's what I posted in response:
```
Hi,

You can load a pairwise distance file either via a 'readdata' command
prior to cluster or by using the 'loadpairdist' option, e.g.

readdata Cmatrix.cmatrix name PW
runanalysis cluster crd1 pairdist PW ...

or

cluster C1 loadpairdist pairdist pw.out ...

If you don't want to do any clustering you can use the 'readinfo'
option to read the results of previous clustering.

I guess it can be a bit annoying if you just want a summary that you
have to specify something to cluster. I never really considered that
case to be honest. If you want, you can "fool" cpptraj by creating a
"fake" data set to cluster that has the same size as the data you
want, then read that in and "cluster" on that after reading in your
pairwise distance matrix. Here's an example where I've modified one of
cpptraj's cluster test cases to do just that.

Original:

# Test loading PW distances from Cmatrix file
cat > cluster.in <<EOF
readdata Cmatrix.cmatrix name PW
parm ../tz2.parm7
loadtraj ../tz2.nc name MyTraj
runanalysis cluster crd1 crdset MyTraj :2-10 clusters 3 epsilon 4.0
summary summary2.dat \
                    complete nofit pairdist PW \
                    cpopvtime normpop.agr normpop
EOF
cpptraj -i cluster.in

The trajectory tz2.nc is 101 frames. So I create a fake data set to
cluster with 101 entries via something like:

#!/bin/bash
rm fakedata.dat
for ((i=1; i<= 101; i++)) ; do
  echo "$i" >> fakedata.dat
done

Then you can use the following modified input: note the replacement of
'crdset' with 'nocoords' and 'data':
# Test loading PW distances from Cmatrix file
cat > cluster.in <<EOF
readdata Cmatrix.cmatrix name PW
parm ../tz2.parm7
readdata fakedata.dat name MyData
runanalysis cluster crd1 data MyData nocoords :2-10 clusters 3 epsilon
4.0 summary summary2.dat \
                    complete nofit pairdist PW \
                    cpopvtime normpop.agr normpop
EOF
cpptraj -i cluster.in

That seems to work just fine. I'll add a feature request to cpptraj
GitHub to make clustering on existing pairwise distance matrices
easier. Thanks for the report!
```

Issue #356, opened 2016-06-17
Make cluster distance calc a separate part of ClusterList


drroe:
Otherwise when reading in info from previous clustering, we can't match representative frames etc from algorithms like HA clustering, which does not use centroid distances to determine cluster distances.

DDGmichigan:
Is this functionality been added to cpptraj? Just wondering?>

drroe:
Addressed by #915

PR #915, opened 2021-09-10
Major overhaul and enhancement of cluster analysis code.

drroe:
Version 6.0.0.

The clustering code has been pretty much entirely rewritten. It was getting to the point where adding new functionality to the old code was becoming prohibitively difficult. 

### Code Changes ###
The clustering code has been reorganized along more logical lines. Clustering consists of an Algorithm class (which drives the actual clustering), a Metric class (which provides the ability to determine centroids, calculate distances between frames/centroids, etc), and various other helper classes (such a classes for each type of centroid). This generalization should hopefully make it easier to add new algorithms in the future, as well as new types of output and cluster-related calculations. The pairwise matrix functionality is now more rational as well; there is a `MetricArray` class containing all Metrics which is responsible for handling all distance calculations (frame to frame and frame to centroid), centroid determination and centroid operations, and the cache for previously calculated pairwise distances. The pairwise cache is a separate DataSet, and is located either in memory or on-disk (eliminating the need for the somewhat nonsensical `DataSet_Cmatrix_NOMEM` class). NetCDF disk cache is now the default.

The new code is in its own namespace and subdirectory, which is probably a tactic I will use more often going forward. The encapsulation of the clustering functionality should make it easier in the future to use this as a backend for e.g. the Wavelet analysis code (which currently has its own DBscan implementation) or for doing something like clustering molecules within a Frame. I've taken pains not to break the interface with pytraj so that should still work, but @hainm and I should talk forward about making the interface more robust (for example, pytraj currently expects a certain ordering to data sets generated by clustering).

### Functionality Changes ###
All functionality is backwards-compatible except for the summary-by-parts output, which now includes for each cluster line a best representative frame calculation for each part (#914). There are 2 major functionality changes:

1. **Any combination of COORDS and 1D sets can be clustered on**, and custom weights can be set for each set. Previously it was either COORDS or a combination of 1D sets. This greatly expands the types of clustering possible. A new keyword, `metricstats <file>` can be used to gain more insight into how each metric is contributing to the total distance.
2. **The `readinfo` keyword can now be used to restart clustering with a different metric/algorithm if desired**. The user can also just specify a cluster number vs time data set for `readinfo` instead of providing an info file. This works for hierarchical aggolmerative, K-means, and DBscan. Other issues that are addressed: #805, #769, #356.

The sieve restore options are no longer locked into specific algorithms and can be chosen by the user. Some potential segfaults were found and fixed (particularly in the DPeaks routine).

Tons of new tests have been added. Also, the manual entry for clustering has been entirely revamped, and missing info (like what SSR/SST is) has been added.

Since this is a gigantic change (and another major version increase) I'm going to label this as WIP for now and let it sit a few days before merging. This has been a long time coming - a few more days won't hurt.


hainm:
> I've taken pains not to break the interface with pytraj so that should still work, but @hainm and I should talk forward about making the interface more robust (for example, pytraj currently expects a certain ordering to data sets generated by clustering).

yeah. `pytraj` uses many hacks to get the data from cpptraj's clustering. :D

drroe:
Added the ability to cluster on any combo of COORDS/1D sets. Added ability to specify weights when multiple metrics are involved. Added `metricstats` keyword for more insight when multiple metrics are involved.

drroe:
Failed Jenkins test is from Linux GNU OpenMP:
```
[2021-09-17T13:22:15.341Z] TEST: /scratch/local/jenkins/cpu/workspace/amber-github_cpptraj_PR-915@2/test/Test_Cluster_SymmRMSD
[2021-09-17T13:22:15.341Z] 
[2021-09-17T13:22:15.341Z]   CPPTRAJ: Clustering with symmetry-corrected RMSD metric (also 2D SRMSD)
[2021-09-17T13:22:16.349Z] terminate called after throwing an instance of 'std::bad_alloc'
[2021-09-17T13:22:16.349Z]   what():  std::bad_alloc
[2021-09-17T13:22:19.366Z] ../MasterTest.sh: line 495:  8991 Aborted                 (core dumped) $cpptraj_cmd >> $CPPTRAJ_OUTPUT
```
I can't seem to reproduce it. `bad_alloc` is usually an out-of-memory thing, but this test should require on the order of 4 MB. I'll try to re-run Jenkins and see if it happens again.

drroe:
run jenkins

PR #929, opened 2022-01-10
Fix segfault when reading certain CHAMBER topology files

drroe:
Version 6.1.1.

The charmm LJ 1-4 terms were being allocated when the FORCE_FIELD_TYPE flag was read; this could cause a segfault if FORCE_FIELD_TYPE came before POINTERS. This PR fixes this behavior by allocating the charmm LJ 1-4 terms (if needed) when they are read.


JaimeRubioMartinez:
My version was 4.25.6
With the last version it work fine !
Thanks

PR #938, opened 2022-01-28
Speed up six-dimensional entropy calculation in GIST

fwaibl:
In GIST, the six-dimensional entropy calculation requires the translational and orientational distance to the nearest neighbor (NN). However, if the translational distance is larger than the 6D distance to the current NN, the orientational distance can be skipped safely.  This is a significant speedup (about 2x in my test), since the orientational distance contains an arccos.

I benchmarked using 10000 frames from a simulation of water around benzene (2700 water molecules). The box size was 80x80x80, with 0.5A spacing. I get the same entropy results, and the time in "Action Post" (from the cpptraj output) reduces from 545 to 263 seconds. I used the OMP version of cpptraj (with CUDA for the energy, but that shouldn't matter).

### DIFF
--- a/src/Action_GIST.cpp
+++ b/src/Action_GIST.cpp
@@ -1326,13 +1326,16 @@ void Action_GIST::TransEntropy(float VX, float VY, float VZ,
     double dd = dx*dx+dy*dy+dz*dz;
     if (dd < NNd && dd > 0) { NNd = dd; }
 
-    int q1 = n1 * 4; // index into V_Q for n1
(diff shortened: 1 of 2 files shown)


drroe:
@fwaibl Just one small thing, could you please increment the revision field in `src/Version.h` by 1 to 6.2.2? E.g.
```C++
#define CPPTRAJ_INTERNAL_VERSION "V6.2.2"
```
Thanks again, this is great!

drroe:
Perfect, thanks!

PR #939, opened 2022-01-31
Faster Hierarchical Agglomerative clustering, output clustering metrics to Data Sets

drroe:
Version 6.2.3.

This PR contains the following improvements.

- Improve the speed of hierarchical agglomerative clustering by an order of magnitude. 
- Create data sets for Davies-Bouldin index, pseudo-F, and SSR/SST ratio clustering metrics.
- Check for and fix a potential integer overflow when setting the random number generator seed from the wall clock.


drroe:
Pytraj failure is because I added 3 new data sets during clustering, causing the offset used by pytraj to access cluster data to be invalid:

https://jenkins.jasonswails.com/blue/organizations/jenkins/amber-github%2Fpytraj/detail/pytraj/414/pipeline#step-22-log-663

@hainm is there any reason why we're using `data[-2]` and not `data[0]` or something? The former (two from the last I think?) means that every time a new data set is added to `cluster` the index becomes invalid. Can we make them `[0]`? I would test myself but I'm having issues getting pytraj compiled on my platform (unrelated glibc issues).

hainm:
> Can we make them `[0]`? I would test myself but I'm having issues getting pytraj compiled on my platform (unrelated glibc issues).

For the record: https://github.com/Amber-MD/pytraj/pull/1598#issuecomment-1026199540

PR #947, opened 2022-02-16
GIST Entropy improvements

fwaibl:
This pull request contains two changes to the GIST entropy calculation:

* Correction term for the "reference volume" in the nearest neighbor entropy. This significantly improves convergence of the entropy in bulk, at the cost of a minor slow-down. There is a flag "oldnnvolume" to recover the old behavior.
* This implements a new version of the nearest neighbor search. The new version allow the user to choose how many layers of neighboring voxels should be used, using a new option "nnsearchlayers". With nnsearchlayers 1 (default), the results are almost exactly the same. Using a higher value improves bulk convergence of the entropy. Furthermore, the new implementation deals differently with waters that are numerically at the same position (this is very rare). 

Timings (Action Post, using 10000 frames of a pure water simulation):
* old, already including the previous speedup: 258s
* with the volume correction: 267 s
* with new NN search (1 layers), and the volume correction: 287 s
* with new NN search (5 layers), and the volume correction: 439 s (all values >1 should have similar timing, since voxels are omitted if it can be proven that the nearest neighbor is already found)

I attach a plot of the average entropy contributions calculated from a pure water MD (lower is better, since the entropy is relative to bulk). Note the log scale.

average-entropy-of-pure-water

In my opinion, the volume correction term is a significant improvement, since it reduces the bulk entropy a lot, and the slow-down is very small. I am not so sure about the new NN search however, since it it less relevant for the convergence (the log scale makes it look bigger, but in fact most of the error is compensated by the volume correction), and the calculation is always a bit slower. I also don't know exactly why it is slower with "nnsearchlayers 1".

Another question: I added some unit tests in the unitTests folder. To do so, I needed to move the new functionality into a separate header (GistEntropyUtils.h) and cpp file. Is the file naming etc. ok with you? I could easily move everything into Action_GIST.cpp, but then I can't use the unit tests any more.


drroe:
> Obviously, the tests are failing because the GIST results are not the same any more. How is this usually handled in cpptraj? Do we need to increase the major (or minor) version number, or do we wait until a new cpptraj version is released?

So typically the way I have been versioning is revisions are for changes which do not change existing output, minor versions are for when input/output changes, and major versions are for when a significant part of the code (API etc) are altered. In this case, we would probably want a minor version bump (resetting the revision number) since output changes (see description in `sec/Version.h`).

> If we need to obtain exactly the same results, I can replace the "oldnnvolume" flag for a "newnnvolume" one and make the old behavior the default. 

Is there any advantage to retaining the old behavior? The new method seems like it is better behaved. However, there is often a benefit to retaining old behavior (for e.g. reproducing old results), particularly if the new results are numerically different. @gosldorf @EricChen521 @tkurtzman what do you think?

@fwaibl if we decide to make this the default, the test results (after being validated) should be updated. If we decide to make it an option, a test should be added for this option. Let me know if this makes sense.

drroe:
@fwaibl also I forgot to mention this, but please run `make depend` in the `src/` directory so that the file dependencies are all updated. Thanks!

drroe:
I'm going to bring your branch up to date. Let me know if it causes any issues.

fwaibl:
I updated the tests and also found a bug meanwhile, where some voxels that are not on the border of the grid were assigned zero entropy values.
I also took the liberty of changing the command line parsing, which was IMO buggy (it was previously assumed that gridcntr is specified before griddim, and the order of the arguments was independent of the keywords - see the commit message for examples.)

src/Action_GIST.cpp
```diff
@@ -231,23 +236,28 @@ Action::RetType Action_GIST::Init(ArgList& actionArgs, ActionInit& init, int deb
...
+    nx = 40;
+    ny = 40;
+    nz = 40;
     mprintf("Warning: No grid dimension values specified, using default (40,40,40)\n");
+  } else {
+    nx = dimArgs.getNextInteger(-1);
+    ny = dimArgs.getNextInteger(-1);
+    nz = dimArgs.getNextInteger(-1);
+  }
   griddim_ = Vec3((double)nx, (double)ny, (double)nz);
```
drroe, line 261: Should exit with an error here if negative dimension values have been specified. I should have done this in the first place.

src/Action_GIST.cpp
```diff
@@ -231,23 +236,28 @@ Action::RetType Action_GIST::Init(ArgList& actionArgs, ActionInit& init, int deb
...
-    gridcntr_[0] = actionArgs.getNextDouble(-1);
-    gridcntr_[1] = actionArgs.getNextDouble(-1);
-    gridcntr_[2] = actionArgs.getNextDouble(-1);
-  } else
+  ArgList centerArgs = actionArgs.GetNstringKey("gridcntr", 3);
+  if ( centerArgs.empty() ) {
     mprintf("Warning: No grid center values specified, using default (origin)\n");
+    gridcntr_ = Vec3(0.0);
+  } else {
+    gridcntr_[0] = centerArgs.getNextDouble(-1);
```
drroe, line 244: I like this change. If we want to be extra careful we can make sure the arguments are valid doubles with the `validDouble()` routine from `StringRoutines.h`, e.g.
```C++
if (!validDouble(centerArgs[0])) {
  mprinterr("Error: '%s' is not a valid argument for 'gridcntr'\n");
  return Action::ERR;
}
gridcntr_[0] = convertToDouble(centerArgs[0]);
```
and so on for the other arguments. Can do something similar for the grid dimensions with `validInteger()`.

src/Action_GIST.cpp
fwaibl: Ok, I can add this quickly. I guess it should also exit if a dimension is zero?

src/GistEntropyUtils.h
```diff
@@ -0,0 +1,88 @@
+#ifndef GIST_ENTROPY_UTILS_H
+#define GIST_ENTROPY_UTILS_H
+
```
drroe, line 3: Ideally we could put all of this in a namespace, e.g.
```C++
namespace Cpptraj {
namespace GistEntropyUtils {
```
I've done this with the clustering routines recently (see the Cluster subdirectory) and I think it makes for nicer code. For some things however it doesn't work since I made `Cpptraj` a class (when I started cpptraj I knew far less about good C++ practices than I do now). I think since only the GIST action will pull this in it would be ok. Let me know what you think.

src/GistEntropyUtils.h
```diff
@@ -0,0 +1,88 @@
...
+#include <cstdlib>
+#include <vector>
+#include "Vec3.h"
+#include <cmath>
+
+typedef std::vector<float> Farray;
+
+#define GIST_HUGE 10000.0
```
drroe, line 11: It would be better to make this a `const double` in a `GistEntropyUtils` namespace, see above.

src/Action_GIST.cpp
drroe: > I guess it should also exit if a dimension is zero?

Yes definitely - I don't think there's any situation where one of the dims can be zero and it would be a valid calculation.

drroe:
@fwaibl I just made a bunch of comments, let me know if you'd like me to handle any of these if you don't have the time. I think I should be able to work on this PR somehow. I can also accept as-is and make the changes in a future PR. Let me know what works best for you.

Thanks for this, it's great work!

drroe:
Also be aware that the code deadline for the next Amber release is March 11, so any changes we want to get into the next Amber  release should be done a few days before that.

fwaibl:
Thanks for your comments. I can add them on Monday, but probably not today. Otherwise, feel free to change something yourself. I think you have write access, since you already pushed a merge commit to my fork (but I am no expert on git...)

> Also be aware that the code deadline for the next Amber release is March 11, so any changes we want to get into the next Amber release should be done a few days before that.

That's good to know. I am confident we can finish this pull request until then, but I probably won't be on time for the extension to ions and multiple solvents, since it turns out that there is quite some stuff that needs to be changed (mostly bookkeeping, i.e., which atoms are solute/solvent, etc.)

drroe:
@fwaibl thanks!

PR #953, opened 2022-03-09
Update GIST section in the manual

fwaibl:
I updated the GIST section in the manual with the two new flags, and also added some information about PME-GIST output and post-processing. I hope I did not mess up any formatting, but the diff seems to be ok. Is it enough to update the cpptraj.lyx file, or do we need to update anything else as well?

Also, I added Johannes Kraml to the contributor list, since he contributed the original GPU implementation of GIST and was also involved with our recent updates to the GIST code. Is this ok?


fwaibl:
Is there a way to avoid all those old commits in the pull request? Everything but the last 3 commits is already merged.

fwaibl:
Sorry, I found a mistake in my update, so I have added another commit. Maybe, someone from Tom Kurtzman's group should read through it.

drroe:
> Is it enough to update the cpptraj.lyx file, or do we need to update anything else as well?

Editing that file should be enough (the Jenkins CI builds the PDF automatically). If you add any references they need to go into the cpptraj.bib Bibtex file. I'll try to build it locally today and give it a look.

> Also, I added Johannes Kraml to the contributor list, since he contributed the original GPU implementation of GIST and was also involved with our recent updates to the GIST code. Is this ok?

Yes, of course! I apologize that it wasn't there to begin with - I try to keep on top of things like that but I sometimes fall way behind.

> Maybe, someone from Tom Kurtzman's group should read through it.

I'll hold off on merging it for now. Just let me know when you'd like it merged. I'll do my best to get it into the Amber tree before the release. Documentation changes are easier to merge in than code changes though so it shouldn't be an issue.

src/Version.h
```diff
@@ -12,7 +12,7 @@
  * Whenever a number that precedes <revision> is incremented, all subsequent
  * numbers should be reset to 0.
  */
-#define CPPTRAJ_INTERNAL_VERSION "V6.4.1"
+#define CPPTRAJ_INTERNAL_VERSION "V6.4.2"
```
drroe, line 15: Since you didnt actually change any cpptraj code you don't need to increment the version. Can leave at 6.4.1.

drroe: I just merged in a PR where I also incremented to 6.4.2 so this should be fine.

drroe:
In the interests of getting this into the Amber tree I'm going to merge this now. Let me know if there are any issues.

Issue #958, opened 2022-03-18
Cpptraj interactive mode can crash out with a long command line


drroe:
Problem is the `CpptrajFile::Printf` buffer gets overrun. Need to either fix `Printf` or use `Write` instead.

drroe:
This should work:
```diff
diff --git a/AmberTools/src/cpptraj/src/Cpptraj.cpp b/AmberTools/src/cpptraj/src/Cpptraj.cpp
index f6857fab2f..5d94b8dacb 100644
--- a/AmberTools/src/cpptraj/src/Cpptraj.cpp
+++ b/AmberTools/src/cpptraj/src/Cpptraj.cpp
@@ -650,8 +650,11 @@ int Cpptraj::Interactive() {
   if (logfile_.IsOpen()) {
     // Write logfile header entry: date, cmd line opts, topologies
     logfile_.Printf("# %s\n", TimeString().c_str());
-    if (!commandLine_.empty())
-      logfile_.Printf("# Args: %s\n", commandLine_.c_str());
+    if (!commandLine_.empty()) {
+      logfile_.Printf("# Args: ");
+      logfile_.Write(commandLine_.c_str(), commandLine_.size());
+      logfile_.Printf("\n");
+    }
     DataSetList tops = State_.DSL().GetSetsOfType("*", DataSet::TOPOLOGY);
     if (!tops.empty()) {
       logfile_.Printf("# Loaded topologies:\n");
```

drroe:
Fixed by #959

PR #959, opened 2022-03-18
Fix memory error when running interactive mode with long command line

drroe:
Version 6.4.4. Address #958. Also corrects some issues highlighted by LGTM.


drroe:
All the jenkins failures were node related. 2 restarts didnt work but no failures are from building or testing so I'm just going to merge.

PR #975, opened 2022-06-23
Minor GIST improvements

drroe:
Version 6.9.0.

In GIST, gracefully handle the case where there is not enough information to calculate entropy (will warn and print 0 now instead of `nan`s). Also keep track of singly-occupied voxels and warn users that more sampling is needed when these are present. (Pinging @fwaibl from our email conversation).

Also provides feedback for `nocom` keyword and updates dependencies.


fwaibl:
Thanks, this looks very reasonable. Maybe we could change the warning to reflect that a few singly occupied voxels might be reasonable? For instance, I checked a GIST calculation with 10000 frames. There, most voxels are occupied around 40 times, but there are still a few voxels around the solute where the population is 1. I assume that this can still happen with higher frame numbers (although less frequently).

drroe:
> Maybe we could change the warning to reflect that a few singly occupied voxels might be reasonable?

Maybe a cutoff, like if > 10% of the voxels have single occupancy? Could make it user-specifiable.

fwaibl:
> Maybe a cutoff, like if > 10% of the voxels have single occupancy? Could make it user-specifiable.

This could be an idea, but I am not sure whether we can find a cutoff that works well across different systems. Maybe we can just make the warning more informative? Something like "%i singly-occupied voxels; To obtain the orientational entropy at those positions, use more frames or bigger voxels."
Generally, I think that this is a nice addition to the code, because it helps the user to find potential problems. But if we print too many warnings, it might be hard to see which ones are the most relevant.

Anyway, I am fine with whatever you decide on. :+1:

drroe:
I went with an informational statement (since you're right, there's nothing that says > X% is bad) and added a percent total:
```
Maximum number of waters found in one voxel for 10 frames = 10
Number of singly-occupied voxels: 166
        166 (13.83 %) of voxels are singly-occupied.
Info: To calculate orientational entropy for these voxels use more frames or bigger voxels.
```

PR #986, opened 2022-08-16
Fix minor bug in PME GIST

drroe:
Version 6.15.0.

When checking if something was on the GIST grid when using PME, the Z comparison was mistakenly using `>-` instead of `>=`. Affects results slightly, test output updated. @fwaibl @tkurtzman @EricChen521


fwaibl:
This is definitely a bug, thanks for fixing it!

Issue #996, opened 2022-11-02
Segmentation fault (core dumped) with GIST  


robinshi123:
Dear all,

I've been running MD simulations using CHARMM and have been trying to run water analysis using GIST in cpptraj.
However. I found that with _griddim_ greater than 10, the calculation always ends up with the same error:
_Segmentation fault      (core dumped)_

If I set the  _griddim_  <= 10, then the calculation is running fine.
I suspect it is caused by memory limitation, so I re-do the calculation on the supercomputer at my university, but I still got the same error. The _griddim_ value on the tutorials seems to be much larger than 10 (~40) so I wonder if it is some issue with the latest version of GIST.

I have attached the input and output files at the end. If you have any thoughts on this, please let me know. Any suggestions would be greatly appreciated. Thanks!

Best,
Jie

cpptraj input file:
```
parm  /scratch/user/shijie0204/sh3/fixed/sh3-cabl/sh3f_cabl_fixed_neu.psf
trajin /scratch/user/shijie0204/sh3/fixed/sh3-cabl/dcd/sh3f_cabl_fixed_md_2.dcd   1 last
gist gridcntr 1 1 2  griddim 11 11 11 gridspacn 0.7
go
```

output in the log file with error:

> CPPTRAJ: Trajectory Analysis. V6.15.1 (GitHub)
>     ___  ___  ___  ___
>      | \/ | \/ | \/ | 
>     _|_/\_|_/\_|_/\_|_
> 
> | Date/time: 11/02/22 18:47:17
> | Available memory: 106.799 GB
> 
> INPUT: Reading input from 'gist.cpptraj.in'
>   [parm /scratch/user/shijie0204/sh3/fixed/sh3-cabl/sh3f_cabl_fixed_neu.psf]
> 	Reading '/scratch/user/shijie0204/sh3/fixed/sh3-cabl/sh3f_cabl_fixed_neu.psf' as Charmm PSF
>     Reading Charmm PSF file sh3f_cabl_fixed_neu.psf as topology file.
> 	PSF contains 13863 atoms, 4375 residues.
> Warning: Determining bond length parameters from element types for 'sh3f_cabl_fixed_neu.psf'.
>   [trajin /scratch/user/shijie0204/sh3/fixed/sh3-cabl/dcd/sh3f_cabl_fixed_md_2.dcd 1 last]
[...]

PR #997, opened 2022-11-04
Add check for non-bonded parameters for GIST

drroe:
Version 6.15.2.

Adds a check for nonbonded parameters for the GIST action. Previously, if using a topology without nonbonded parameters, cpptraj would segfault. This adds an informative error message, and also adds a warning if there is no charge information.


drroe:
This should address #996

PR #999, opened 2022-11-16
Add NetCDF4/HDF5 support, including support for compressed NetCDF trajectories.

drroe:
Version 6.16.0.

This PR adds support for the NetCDF4/HDF5 format in addition to the current NetCDF 3 (classic) format currently used by Amber. The main advantage of this is it makes it possible to natively compress NetCDF trajectories via the underlying HDF5 framework. There are currently two options for compressing NetCDF trajectories; lossless, and lossy; the latter uses integer quantization to improve compression ratios and is explained in an upcoming paper to be published in Protein Science.

- Support NetCDF4/HDF5 formats (only serial currently) - `hdf5` keyword for NetCDF `trajout`.
- Support compression and lossy compression of NetCDF4/HDF5 formats - `compress`/`icompress` keywords for NetCDF `trajout`.
- Support reading (not writing) MDtraj H5 and MDAnalysis H5MD trajectories.
- Add `nokinetic` keyword to the `energy` action to suppress kinetic energy calculation when velocity information is present.
- Can now read Amber energy files (`amberene`).
- Add `prec {single|double}` keyword to `loadcrd` command to specify saving coordinates as single (COORDS) or double (FRAMES) precision.
- Add `bondparm` keyword to the `change` command for modifying bond parameters.
- Add concatenation of string data sets with `dataset cat`.

Manual has been updated and tests added.

Still left to do:
- NetCDF4 files cannot be written in parallel yet; they require slightly different code than classic (which happens via `pnetcdf`).
- The `cmake` build does not yet know how to link in HDF5.

Other changes in this PR:
- Fix error detection in `BufferedFrame::ReadFrame()`.
- Fix error checking for Gzip IO (`FileIO_Gzip`).
- Fix minor bug in the `ArrayToRangeExpression()` routine that could cause it to incorrectly fail (and also cause a minor memory issue).
- New command `compareenergy` for comparing individual components of energy calculations. Still hidden, not ready for release yet.
- Add `Rattle()` routine to `Constraints` class for potentially doing positions and velocities (not used yet).


drroe:
Note: Appveyor failures were due to updates on the Appveyor platform itself. Reverting to the previous VS 2019 image fixed the build issues.

PR #1001, opened 2022-11-18
Add missing files to static libs

drroe:
Related to #999. No version bumps needed, only affects build framework.


drroe:
I'm getting tired of keeping appveyor going. Error is now because suddenly netcdf is not being found. Might be time to cut that build loose.

PR #1005, opened 2022-12-02
Add ability to give separate data set and file names for cluster pairwise matrix

drroe:
Version 6.16.2.

Adds a new keyword to the `cluster` command, `pairdistfile`, which can be used to specify the pairwise cluster matrix file name. Previously `pairdist` was used to set both the data set name and file name. This is still the behavior if `pairdist` is specified and `pairdistfile` is not.

Updated the manual and added a test.


drroe:
LGTM is shutting down this month - the failure may be related.

PR #1015, opened 2023-02-21
Fix xtalsymm command use of unit cell matrices

drroe:
Version 6.17.4. 

Previously, the `xtalsymm` command was using the unit cell matrix to perform rotations, but what was needed was the transpose (inverse rotation). Also adds some more useful outputfor the `xtalsymm` command (explicitly listing the detected asymmetric units).


drroe:
@swails @hainm why did this iteration of pytraj pass I wonder? What's different in the python setup?

hainm:
I don’t really know 😅.

PR #1019, opened 2023-03-10
Miscellaneous cleanup

drroe:
Version 6.18.1.

- Disable the base pair Z axis `[nxyz]` output from `nastruct` by hiding it behind the `NASTRUCTDEBUG` define for now since pytraj doesn't yet know how to deal with vector data from the `nastruct` command (see https://github.com/Amber-MD/pytraj/issues/1624). Also remove the reference to `nxyz` from the `nastruct` manual for now.
- To be more thorough, add some `distance` tests and a `vector minimage` test that actually require some imaging corrections.
- Disable the appveyor cmake mingw test since `pacman` on their VS 2015 image seems hopelessly broken. Try to get the configure mingw test to build its own libraries. If that doesn't work, that test will be disabled as well since it's not worth maintaining.


drroe:
Note: this will likely be the version released with AT 23.

PR #1025, opened 2023-04-26
Update version of NetCDF that CPPTRAJ will try to build.

drroe:
The NetCDF 4.6.1 TAR file appears deprecated. Use a newer NetCDF TAR file (4.9.2). Disable CURL support for the bundled NetCDF to simplify linking.


drroe:
Jenkins failure is the spurious pytraj one.

PR #1027, opened 2023-05-10
Fix unwrapping and diffusion imaging for NPT trajectories

drroe:
Version 6.19.0.

Previously, unwrapping and imaging for the diffusion calculation were done in a manner that became very inaccurate for long time scales. Now unwrapping and the imaging for the diffusion calculation are done entirely in fractional space, which largely eliminates this problem.

Also correctly read time information from DCD trajectories, and adds 3 new keywords for controlling time info written to DCD trajectories:
```
	dt      : Set trajectory time step in ps.
	nstep   : # steps between frames.
	step0   : Initial step.
```


drroe:
Jenkins failure is the spurious pytraj one. Ignoring it.

PR #1028, opened 2023-05-11
Add ability to remove box fluctuations when unwrapping NPT trajectories.

drroe:
Version 6.19.1.

Adds new command for calculating average unit cell:
```
  [help avgbox]
	[<setname>] [out <file>]
  Calculate average box.
```
This in turn can be used with the `unwrap` command for removing box fluctuations from NPT trajectories when unwrapping via the new `avgucell` keyword. Adds tests for both, updates manual (including an example and advising people to do this when calculating diffusion for NPT trajectories).


drroe:
Usual pytraj jenkins failure.

PR #1032, opened 2023-06-08
Add diffusion calculation from multiple time origins and toroidal-view scheme

drroe:
Version 6.19.3. Adds new commands and keywords related to the calculation of diffusion constants.

1. Add the `calcdiffusion` Analysis command.
2. Add the `tordiff` Action command.
3. Add the `scheme` keyword to the `unwrap` Action command.

The `calcdiffusion` analysis command will calculate diffusion from multiple time origins.
```
	[crdset <coords set>] [maxlag <maxlag>] [<mask>] [time <dt>]
	[<name>] [out <file>] [diffout <file>]
```
This command does not perform imaging, so coordinate need to be unwrapped prior. This is the first Analysis command to be MPI parallelized (and is also OpenMP parallelized).

The `tordiff` action command calculate diffusion using the toroidal-view-preserving scheme (https://arxiv.org/abs/2303.09418), which is an alternate way of unwrapping NPT trajectories. It only works for orthogonal cells currently.
```
[<set name>] [<mask>] [mass] [out <file>] [diffout <file>] [time <dt>]
```

The `scheme` keyword for the `unwrap` Action (`[scheme {frac|tor}]`) will optionally use the toroidal-view-preserving scheme for unwrapping instead of the default of fractional coordinates. This currently only works by molecule and for orthogonal unit cells.

Updates the manual and also adds tests.

### DIFF
--- a/src/Analysis_CalcDiffusion.cpp
+++ b/src/Analysis_CalcDiffusion.cpp
@@ -169,7 +169,7 @@ Analysis::RetType Analysis_CalcDiffusion::Analyze() {
     return Analysis::ERR;
   }
   if (TgtTraj_->Size() == 1) {
-    mprinterr("Error: COORDS set '%s' has only 1 frame.\n");
+    mprinterr("Error: COORDS set '%s' has only 1 frame.\n", TgtTraj_->legend());
     return Analysis::ERR;
   }
   if (TgtTraj_->Top().SetupIntegerMask( mask1_ )) {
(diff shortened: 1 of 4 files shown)


drroe:
@swails @hainm if you get a chance can you see if you can reproduce this pytraj error: https://jenkins.jasonswails.com/blue/organizations/jenkins/amber-github%2Fpytraj/detail/pytraj/599/pipeline

PR #1034, opened 2023-06-16
Fix parallel analysis when ensemble is active

drroe:
#1032 changed previous behavior for analysis in parallel when `ensemble` has been used; ensemble masters beyond the first member were no longer performing analysis. This PR restores the old behavior. Also adds a test for this, enables the `calcdiffusion` test, and fixes the clean rule for the GIST test.


drroe:
Jenkins timed out in the post-test stage. 🤷‍♂️ All the builds passed, but the manual and pytraj stages didn't execute. No idea what's up.

PR #1035, opened 2023-07-07
Try to add a pytraj GitHub Actions build job

drroe:
Testing pytraj with cpptraj on Jenkins has been unreliable. Try to use GitHub actions instead.


drroe:
The holdup is Jenkins:
```
All nodes of label ‘pgi&&intel-cpu’ are offline
```
Given that this will probably just time out anyway, pytraj is no longer tested there, and all the other tests have passed, I'm going to go ahead and merge this. The only (small) downside is that the manual will not be built (removed a duplicate entry in a table).

PR #1042, opened 2023-07-21
Add 'openmm' keyword to 'energy' action

drroe:
Version 6.20.2. Add 'openmm' keyword to 'energy' action to use OpenMM (if support compiled in) to calculate total energy.


drroe:
Not sure why the pytraj build fails - it works locally. Will try to re-run.

drroe:
I hate python. Keeping python CI running is whack-a-mole nonsense.

hainm:
> I hate python. Keeping python CI running is whack-a-mole nonsense.

:D

swails:
> I hate python. Keeping python CI running is whack-a-mole nonsense.

I think pytraj may be a little over-tested.  But the two real solutions I see here are:

1. Find a proper maintainer for pytraj.
2. Deprecate pytraj and push people towards alternatives like MDAnalysis or mdtraj.

I know pytraj *is* used (some people use it here at Entos), so the 83 "dependents" reported by GitHub is certainly an underestimate, and possibly significantly so.

Judging from my own experience with ParmEd, @hainm does not have the bandwidth to maintain pytraj (in my case, needs here at Entos now afford me to expend some time on ParmEd maintenance and development, but that was not always the case).

---------

@drroe or @hainm - proposals or suggestions here?  As it stands right now, pytraj is an unfair weight to tie around cpptraj development.

drroe:
First of all, I appreciate the in-depth and earnest commentary from @swails and @hainm - I thought my comment was just a little screaming into the void and didn't expect a response (although the conversation is certainly worth having).

> If he gets frustrated enough with the cost of keeping pytraj working and decides to stop doing it and turn off the tests, then the end result is the deprecation of pytraj

I may not love developing in python, but I recognize its utility for users and I think pytraj is worth maintaining.  @hainm has put a lot of work into it and I don't think it has run its course yet.

> Basically, if pytraj lacks a willing maintainer, the choice of deprecation falls to whoever is left holding the bag

This is an important point, and perhaps worth a short virtual (zoom/hangouts) extended conversation if you two are up for it?

PR #1052, opened 2023-10-13
Add ability to scale charges with the 'graft' command

drroe:
Version 6.20.6.

Add 2 new keywords to the `graft` command, `srccharge` and `tgtcharge` which allow the charges of the source/target fragments to be scaled to a user-specified value. Updates manual, adds tests.


drroe:
Jenkins PGI build is hanging. I'm going to ignore it.

PR #1053, opened 2023-11-01
Add ability to utilize and manipulate internal coordinates

drroe:
Version 6.21.0.

Add commands/keywords related to internal coordinates to cpptraj:

- `zmatrix` : Apply Z-matrix to a COORDS set or calculate Z-matrix for a molecule/frame in a COORDS set.
```
  [help zmatrix]
        <COORDS set name> [name <output set name>]
        { zset <input zmatrix set> [parm <top>|parmindex <#>] |
          [molnum <mol#>] [frame <frame#>] [out <zmatrix file>] }
```

- `sequence` : Create a new molecule from a sequence of COORDS sets.
```
 [help sequence]
        name <output set name> <unit0> <unit1> ...
        [{libset <libsetname>} ...]
```

- `ic` keyword for the `graft` command: connect source and target without having to RMS-fit.

Also adds the ability to read in units from Amber Prepin and OFF library files as COORDS data sets; the latter can be used directly with the `sequence` command. Updates the manual and adds new tests. Lots of additional fixes under the hood.


drroe:
Sometimes CI runners are no fun. New issue popping up that has never been a problem before - clang/gfortran incompatibility:
```
Error: Test compile failed: clang++  -Wall  -O3     -std=c++11 -o testp testp.cpp /home/runner/work/cpptraj/cpptraj/lib/liblapack.a /home/runner/work/cpptraj/cpptraj/lib/libblas.a -lgfortran -lpthread    
  Checking built LAPACK/BLAS: 
Error: Check the output below for error messages:
/usr/bin/ld: cannot find -lgfortran: No such file or directory
```
Not sure how to fix it yet.

drroe:
Turns out I had already fixed this transient issue (clang/gfortran) in `configure`, but only for OSX. I just took that code and made it apply to all OS - seems to have done the trick. 🎉

PR #1059, opened 2023-12-12
Add note about strip and atomicfluct with `pdbres` keyword

slochower:
Added two quick notes about `parmopts` and `pdbres` options for `strip` and `atomicfluct`.


drroe:
Thanks for doing this!

drroe:
Jenkins and the pytraj CI stuff is currently broken so those arent real failures.

drroe:
I'll update this with the latest PR and merge when the CI is finished.

PR #1063, opened 2024-01-23
If wildcard expansion fails in a list for loop, do not add to the loop.

drroe:
Version 6.22.4. Fixes an incompatibility with the [AmberMdPrep.sh](https://github.com/drroe/AmberMdPrep) script. 
```
for FILE in final.?.out,final.??.out
  readdata $FILE name MD
done
evalplateau *[Density] name EQ out Eval.agr resultsout Eval.results
```
The evaluation portion would fail if no `final.??.out` files are present. This PR fixes that behavior (restores original behavior).


drroe:
PGI machine is down again. Merging.

PR #1068, opened 2024-03-06
Do not use "bad" box info from PDB files

drroe:
Version 6.24.0.

Some PDBs (like 1FHK) contain `CRYST1` records where the unit cell lengths are all 1.0 Angstrom; this should be ignored (it even says so right in the PDB itself):
```
REMARK 215 NMR DATA.  PROTEIN DATA BANK CONVENTIONS REQUIRE THAT         
REMARK 215 CRYST1 AND SCALE RECORDS BE INCLUDED, BUT THE VALUES ON      
REMARK 215 THESE RECORDS ARE MEANINGLESS.
```
This PR changes cpptraj's behavior so that when this "bad" unit cell info is encountered it is ignored, which prevents bad imaging from taking place. As a result cpptraj will no longer print the bad CRYST1 records.

### DIFF
--- a/.github/workflows/merge-gate.yml
+++ b/.github/workflows/merge-gate.yml
@@ -148,7 +148,10 @@ jobs:
         echo $CONDA/bin >> $GITHUB_PATH
     - name: Install conda packages
       run: |
-        $CONDA/bin/conda env update --file devtools/ci/environment.yml --name base
+        which conda
+        conda install conda=23.11.0 python=3.10
+        conda --version
+        conda env update --file devtools/ci/environment.yml --name base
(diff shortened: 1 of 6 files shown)


drroe:
GRRRR why is it that when there is an error in the CI framework it's _almost always PYTHON-related_ and unrelated to any code changes. I hates it precious...😒

drroe:
Jenkins doesn't seem to be running, or the pipeline isn't launching for some reason.

drroe:
I've confirmed (locally) this works with OneAPI 2022.2 so I think it's good to merge.

PR #1086, opened 2024-06-13
CIF fixes, preserve CHARMM PSF segment ID

drroe:
6.27.0. Allows CPPTRAJ to handle chain IDs that are larger than 1 character, which is possible with the MMCIF format. As a bonus this now allows CHARMM PSF segment IDs to be preserved. Detect "bad" unit cell info in CIF files (i.e. all lengths are 1.0 Ang. like in PDBs), which will fix improper imaging of CIFs with this kind of unit cell info.

Updates existing PSF test.


drroe:
@hainm I have a fix for pytraj that I will push once this is merged.

PR #1088, opened 2024-06-26
Handle blank lines at the end of Karplus data file for jcoupling action

drroe:
6.27.1.

Fixes behavior where CPPTRAJ would segfault if there were blank lines at the end of the Karplus data file in the `jcoupling` action.

### DIFF
--- a/src/Action_Jcoupling.cpp
+++ b/src/Action_Jcoupling.cpp
@@ -56,9 +56,16 @@ int Action_Jcoupling::loadKarplus(std::string const& filename) {
   if (buffer == 0) return fileEOF(filename);
   while (buffer != 0) {
     // Skip empty lines (BufferedLine.Line() removes newlines) and comments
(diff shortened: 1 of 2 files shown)


drroe:
@hainm any insight into the pytraj error here: https://github.com/Amber-MD/cpptraj/actions/runs/9685730413/job/26726487745?pr=1088

Could it be related to the recent changes? As far as I can tell this change shouldnt affect pytraj at all.

hainm:
I think it does. (Guess that from reading the title).
I will confirm this later (on the way home)).

Hai

hainm:
> I think it does. (Guess that from reading the title). I will confirm this later (on the way home)).

Oh, my mistake. It's because I did code refactoring and the code uses `strenum`

I will fix this.

hainm:
green now

image

drroe:
Thanks Hai!

Issue #1090, opened 2024-06-28
Res TYR:1063 is missing atoms N


H-EKE:
Hi ,

I am trying to calculate the secondary structure of a protein using charmm36

```
parm prod1/structure.psf
for i=1;i<5;i++
  trajin prod$i/output.xtc 1 5000
  strip :TIP3,SOD,CLA,POPC
  secstruct :1029-1063 nameh HN out ss_per_res_y_$i.dat sumout dssp_y_$i.dat totalout total_y_$i.out
go
done
```

When I run that script I get the following output

```
CPPTRAJ: Trajectory Analysis. V6.4.4 (GitHub) MPI OpenMP
    ___  ___  ___  ___
     | \/ | \/ | \/ | 
    _|_/\_|_/\_|_/\_|_

| Running on 1 processes.
| 1 OpenMP threads available.
| Date/time: 06/23/24 11:20:51
| Available memory: 266.581 GB

INPUT: Reading input from 'STDIN'
  [parm prod1/structure.psf]
        Reading 'prod1/structure.psf' as Charmm PSF
    Reading Charmm PSF file structure.psf as topology file.
        PSF contains 570230 atoms, 145435 residues.
Warning: Determining bond length parameters from element types for 'structure.psf'.
  [for i=1;i<5;i++]
    Setting up 'for' loop.
CONTROL: Parsing control block.
  BLOCK  0: for ((i=1; i<5; i+=1)) do
  [trajin prod$i/output.xtc 1 5000]
        Added command 'trajin' to control block 0.
  [strip :TIP3,SOD,CLA,POPC]
        Added command 'strip' to control block 0.
  [done]
  BLOCK  0: END
CONTROL: Executing 1 control block(s).
        Loop over 'i' will execute for 4 iterations.
  [trajin prod1/output.xtc 1 5000]
        Reading 'prod1/output.xtc' as Gromacs XTC
Warning: Trajectory box type is 'Orthorhombic' but topology box type is 'None'.
Warning: Setting topology box information from trajectory.
  [strip :TIP3,SOD,CLA,POPC]
    STRIP: Stripping atoms in mask [:TIP3,SOD,CLA,POPC]
  [trajin prod2/output.xtc 1 5000]
        Reading 'prod2/output.xtc' as Gromacs XTC
  [strip :TIP3,SOD,CLA,POPC]
    STRIP: Stripping atoms in mask [:TIP3,SOD,CLA,POPC]
[...]

PR #1091, opened 2024-07-02
Fix which atoms are reported in DSSP "missing atoms" warning message

drroe:
6.27.2. Related to issue #1090. When reporting missing atoms, cpptraj was mistakenly reporting the `N` atom when `O` was missing and vice versa. This PR fixes the problem.


drroe:
@swails Jenkins build cannot finish:
```
Queued: All nodes of label ‘docker&&intel’ are offline
```
All other tests appear to have passed.

Issue #1070, opened 2024-03-15
Problems on ARM systems


drroe:
Reported on the Amber mailing list on 3-15-2024 by user Hirosuke Hotta:
```
Dear cpptraj developers,

Regarding cpptraj error with test "Test_SPAM".
We found Test_SPAM (there are 2 tests in RunTest.sh, what I say here is the 1st one) was abnormally ended with segmentation fault on the ARM version of RHEL8.8, which was not seen on the ARM versions of RHEL8.7 or older.
We also found the similar symptoms were recognized on some versions (arm64, riscv64, sparc64 and so on) of Debian Linux (https://tracker.debian.org/pkg/cpptraj).

After investigation, we found the segmentation fault occurred at "vsprintf(linbuffer_,format,args);" in the function "CpptrajFile::Printf" (CpptrajFile.cpp).
In the function,
- The value of "format" was "%8.2147483647f" on ARM Linux system, while "8.3f" on Intel Linux system
- The value of "linebuffer_" is, we think, 1024 or so, which is much smaller than 2147483647
- On some ARM system (like RHEL8.7), Test_SPAM works, BUT summary.dat looks anomalous, i.e., we don't see any numbers in the column of "#Peak"
- On some ARM system (like RHEL8.8), Test_SPAM fails with the segmentation fault.
- As far as we know, this problem may occur with AmberTools 20, 21, 22 and 23.

Also we found the fundamental causes were as follows. In the function "FloatWidth" (StringRoutines.cpp),
- there are no checks with the argument for log10() in the code, and unfortunately 0.0 is passed as the argument, and
- the return value "(int)float_exponent" (= (int)(fabs(log10(0.0))+1)) is different between ARM and Intel
    on ARM:    2147483647 (max value of signed int)
    on Intel: -2147483648 (min value of signed int)

For our idea of modification and the detail of what happens, please see the attached file.

Regards,
Hiro
```

drroe:
This should hopefully be addressed by #1093. If you get a chance to try it out let me know if it works for you now. Thanks again for the report!

PR #1093, opened 2024-07-19
Fix compilation on ARM based systems

drroe:
Version 6.28.0.

This should fix the test failures on ARM systems observed in #1070 (I finally got access to an ARM system for testing). The problem was simple; the function that is supposed to detect the number of leading zeros in a floating point number fractional part wasn't properly protected against a value of 0 being passed to `log10()`. On ARM systems this failure lead to bad precision values which in turn gave bad output to ASCII data files. This PR fixes the behavior.


drroe:
Build of the manual is still broken. 😢

PR #1099, opened 2024-08-29
Recognize Drude particles, improve handling of lone pairs

drroe:
Version 6.29.0.

- Drude particles are now identified from Charmm PSF files and given the element type `DR`. 
- The lone pairs section is now properly read from Charmm PSF files.
- The `distance`, `angle`, `dihedral`, and `vector` actions now check if the total mass of atoms in a mask is zero when center of mass is requested; this can happen when e.g. only lone pairs are selected.
- Add the `geom` keyword to `vector` to enable vector calculation using geometric centers instead of center of mass (to get around the aforementioned zero mass problem).
- Add the `debye` keyword to `vector dipole` to report dipole vector in units of Debye. 
- Lone pairs are given different (shorted) default distances when no bond parameters are present.
- Add the `bondparm` command for printing the bond parameter table.

Adds a test for reading Charmm PSF files with Drude particles. Updates the manual.


hainm:
I will have a look next week, I will have an international flight.

Hai

hainm:
> Grr, python 😠 @hainm can you take a look? https://github.com/Amber-MD/cpptraj/actions/runs/10618137181/job/29432643446?pr=1099
> 
> I'm going to merge this stuff if everything else passes this seems like one of those "with some python versions it works, with some it fails" errors, i.e. the errors are not related to the changes in this PR.

 @drroe It seems like a real issue instead of python thing. um. I don't know why yet.

image

drroe:
@hainm looks like the problem is `pytraj` isn't fully compatible with numpy 2.1.0. When I pin numpy to 2.0.0 the tests pass again (see https://github.com/Amber-MD/cpptraj/pull/1102).

hainm:
> @hainm looks like the problem is `pytraj` isn't fully compatible with numpy 2.1.0. When I pin numpy to 2.0.0 the tests pass again (see #1102).

yay. thanks for resolving this in cpptraj. I will think about a solution for numpy 2.1.0.
I think it's because of Cython version that generates the c++ files.

PR #1100, opened 2024-08-30
Minor fix for PSF with NUMLP

drroe:
6.29.1. Previous PR made it so that PSF read could fail when no NUMLP section present. This PR fixes this. Also hides some leftover debug info.


drroe:
Python version weirdness again, failure not related to cpptraj code changes.

PR #1114, opened 2024-11-07
Allow atommap rmsfit to proceed with partial atom maps

drroe:
6.29.8. Previously `atommap rmsfit` required a full atom mapping to work. This PR changes it so that only the mapped atoms are used in the rms fit. It also adds a check that there are at least 3 mapped atoms for this, and checks for `rmsfit` and `changenames` (the two are mutually exclusive).

### DIFF
--- a/src/Action_AtomMap.cpp
+++ b/src/Action_AtomMap.cpp
@@ -56,6 +56,10 @@ Action::RetType Action_AtomMap::Init(ArgList& actionArgs, ActionInit& init, int 
   maponly_ = actionArgs.hasKey("maponly");
   rmsfit_ = actionArgs.hasKey("rmsfit");
   renameAtoms_ = actionArgs.hasKey("changenames");
+  if (rmsfit_ && renameAtoms_) {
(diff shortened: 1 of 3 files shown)


hainm:
@drroe how is Jenkins used nowadays in cpptraj?

drroe:
> @drroe how is Jenkins used nowadays in cpptraj? It seems the build keeps failing in this.

The failure was unusual (a `std::bad_alloc`) and I have made some under-the-hood changes to the `Frame` class in this PR, so this could very well be a legit failure. I haven't had the time to work on it lately though which is why it is currently in limbo.

hainm:
@drroe how's about this https://github.com/Amber-MD/cpptraj/pull/1125?

drroe:
> @drroe how's about this #1125?

It's a real failure:
```
ReadLine.cpp:11:13: fatal error: readline/readline.h: No such file or directory
```
Caused by the changes to the readline includes needed to get cpptraj working with the conda readline. It will require me reworking how the bundled readline is included (I think). I'm testing that in a separate branch.

PR #1126, opened 2025-01-06
Try to improve compatibility with system readline libraries

drroe:
Change the way `readline` is included so that it is consistent between the internal `readline` and system `readline`s. Also add better checks for whether `ltermcap` or `lncurses` is needed. Should help the conda package.


drroe:
```
python 3.10**  is not installable because it conflicts with any installable versions previously reported.
```
Have I mentioned before how much I detest python's versioning hell...

PR #1150, opened 2025-09-18
Docs: updated manual with small edit

rodrigogalindo:
Small edit in the CPPTRAJ manual

PR #1151, opened 2025-09-20
Add README for documentation.

drroe:
Adds a README describing the files in the documentation subdirectory `doc`.

@rodrigogalindo this is also being used as an independent test of #1150 to see if the Appveyor/Jenkins test failures can be replicated.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -1,0 +1,8 @@
+Cpptraj Documentation
+=====================
+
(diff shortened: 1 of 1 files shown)


drroe:
The failing appveyor mingw tests are all related to FFT. I wonder if the recent update from mingw 14 to mingw 15 is the cause. Either mingw 15 is broken for the fortran-based PubFFT, it's not compiling it correctly because some flag(s) isn't set, or there is some long-hidden bug in PubFFT that is being exposed. Might be worth it to try to get pacman to download fftw.

Issue #1187, opened 2026-07-08
Segfault when doing trajin of certain single trajectory frames followed by another trajectory.


drroe:
To reproduce:
```
parm tz2.parm7
trajin tz2.rst7
trajin tz2.nc
trajout seg_stripped.nc
go
quit
```
Confirmed if the first trajectory is Amber restart, NetCDF restart, NetCDF traj, TRR. Does not happen with mol2 or DCD so far.

drroe:
The problem is when a trajectory with velocities and or forces is followed by a trajectory without velocities/forces. The output trajectory is initially set up to expect velocities/forces, so for the subsequent trajectory without that info the pointer is null. The reason the opposite way works is (I think) that each Traj_XXX class is set up to ignore forces/velocities in the incoming frame if the initial trajectory was not set up with them.

3 possible solutions:

1. Modify SetupEnsembleOut()/SetupTrajout() and their parallel equivalents (called from CpptrajState) to check for changed CoordinateInfo. Will require passing in the Frame(s) for possible modification.
2. Create a new function that does this check/modification, called from similar spots in CpptrajState.
3. Modify each Traj_XXX class to protect against this.

I'm leaning towards 2 or 3. 2 will make it so that warnings are easier and future trajectory classes will automatically be protected. 3 is good because they should be protected for both cases. Maybe I need to do a combination of the two, even though that's a lot of work.

drroe:
Should be addressed by #1188

PR #1188, opened 2026-07-15
Fix segmentation faults when reading in trajectories without forces/velocities after reading in trajectories with forces/velocities

drroe:
Version 7.10.0. Should address #1187.

In the case where loading in a trajectory with velocity/force information before a trajectory without velocity/force information, the output trajectory would be set up with that information expected. When encountering a trajectory without that information a segfault would occur. This PR fixes the behavior so that zeroes for forces/velocities are written instead of segfaulting when this occurs.  This is fixed for both `trajin` and `ensemble` in serial and in parallel.

Adds a check that replica indices map to the same values when loading in multiple ensembles; otherwise sorting may not work properly.

Adds 3 new math functions: hyperbolic sin, cos, and tan.

Adds a new keyword for `ensemble`,  `bytemp` to force sorting by replica temperature (instead of replica indices).

Adds more tests and updates the documentation.


drroe:
According to `git bisect`, the first bad and last good commits:
```
commit 066c404e430670a427abe7d694e02f88802a673b (HEAD)
Author: Daniel R. Roe <daniel.r.roe@gmail.com>
Date:   Tue Jul 14 16:12:44 2026 -0400

    Version 7.10.0. Minor version bump for fixing handling remd ensembles with changing vel/frc info. Also added 3 math funcitons.

commit 784e9fd9fd0aa6a0c3abaa48bd8739082a86cf2a
Author: Daniel R. Roe <daniel.r.roe@gmail.com>
Date:   Tue Jul 14 15:40:43 2026 -0400

    Hide debug info
```

drroe:
Sometimes I am a complete dingdong...

Need to use NcTest() to compare NetCDF files, not DoTest() :man_facepalming: . The version bump was causing the files to look different.

drroe:
The Jenkins failure is a non-specific one for the CUDA build. Potentially the CUDA runner is having issues. Local CUDA compiles and tests fine so I am going to merge.

PR #1192, opened 2026-09-22
Fix segmentation fault when stripping topologies containing LJC parameters

drroe:
Version 7.11.2. The LJC array was not being properly resized on the fly when stripping a Topology containing LJC parameters. This PR fixes this.


drroe:
All Jenkins nodes are currently offline.

